PR #2, opened 2021-10-05
readme: add draft of npx instructions

e-nastasia:
Just a small draft adding another method of using the scaffolding tools. Instead of writing this command somewhere else, we can keep all the docs close to code :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,18 @@ RAD tools to enable quick scaffolding of Holochain application.
 
 # Usage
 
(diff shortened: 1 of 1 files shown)


guillemcordoba:
The thing is: nobody will need to use npx in the future, when this stabalizes. `npm init @holochain` is what everyone will be supposed to be doing. We had to use the npx because it enables us to pin-point the exact version of the scaffolding tools we wanna use, and npm init @holochain does not (specially when re-running a second time). Does that make sense?

PR #5, opened 2022-01-24
Update README.md with nix-shell instruction

bierlingm:


guillemcordoba:
@bierlingm could you please describe the scenario in which you had any problems/issues with the scaffolding tool? The idea is that devs that don't even have nix installed can also run `npm init  @holochain` directly.

bierlingm:
> @bierlingm could you please describe the scenario in which you had any problems/issues with the scaffolding tool? The idea is that devs that don't even have nix installed can also run `npm init @holochain` directly.

Yeah, I don't anymore remember what the problem was I had, just that I had one. But I attempted the regular usage with `npm init @holochain` and it worked perfectly fine. Please disregard. :D

PR #82, opened 2023-03-28
updates hello-world

zippy:
Updates example hello world app to:

- simplify validation
- add a hello entry type with send-hello
- add UI to get hellos from DHT and render


src/scaffold/zome/coordinator.rs
```diff
@@ -55,8 +55,20 @@ pub fn initial_lib_rs(dependencies: &Option<Vec<String>>) -> String {
             s
         }
     };
-
-    format!(
+    if simplified {
```
guillemcordoba, line 58: Do we really need this? Are you concerned about them doing `hc scaffold zome` in an app where they have used the vanilla template, so after they have scaffolded an example hello world? If not, I don't see what effect this has on `hc scaffold hello-world` itself.

guillemcordoba (commented):
Just a comment, but mostly looking good and yes this is the way I would have done it. 

Maybe in the index.html I would have polled the hellos instead of having a button to fetch them but that's mostly up to style and what we think that will fit the beginner devs' frame better.

src/scaffold/zome/coordinator.rs
zippy: Oh right, that was my old stuff before we went down the route of moving it all to the template.  I will remove.

zippy: Fixed

zippy:
I like the idea of the button click because I want an interactive experience in the hello-world.

guillemcordoba (approved):

PR #85, opened 2023-04-05
Add holo

robbiecarlton:
This adds a(n undocumented) `--holo` flag to the `web-app` subcommand.

When you pass that command `--holo`, it generates a hybrid launcher/holo UI that can be run in either mode, together with npm scripts to run it in either mode.

One important caveat:

The holo side of these npm scripts depends on `holo-dev-server`, a binary that allows you to spin up a local holo dev environment for manual (and automated) testing.

The current solution for getting this binary to people is not satisfactory, and we're going to update it soon, but what that means is I haven't built in any way to get that binary from the holo generated happs yet. So this PR just assumes you have access to it locally, and if not, the holo npm scripts won't work.

Obviously this is not ideal.

Possible ways forward, from most to least conservative:

1) Keep this PR unmerged until we've answered the `holo-dev-server` distribution question. I'll just keep an eye on the repo and make sure to keep this up to date
2) Build some kind of warning into the npm script so that if you run it, it tells you that you're missing holo-dev-server and points you in the right direction
3) Add per platform `get-holo-dev-server` scripts that users can run to curl it down. This will be limited to the couple of platforms we currently support, (x86 linux and mac).  Outside of that, users will have to wait til we've answered the distribution question
4) Just merge it as is, trusting that the only people using it while it's undocumented will already have / know how to get `holo-dev-server`

@guillemcordoba - 1) is the most labor intensive for me, but shouldn't be too bad, and I'm happy to go that way if you don't feel comfortable with any of the others.

lmk if you have any questions


guillemcordoba (approved):
The changes look perfect to me, just what I would have done as well. Mmm I like the options you laid out, and actually I think I'd prefer merging this in, as this option will be hidden by default. Do you have an idea of how getting the holo-dev-server will look like for mac and linux? Maybe this is something to talk to @steveeJ about, to integrate holo-dev-server into holonix.

robbiecarlton:
thanks @guillemcordoba! Dm'd you about the holo-dev-server question

PR #86, opened 2023-04-17
Bump to holochain v0.1.4

robbiecarlton:


src/scaffold/dna.rs
```diff
@@ -216,7 +216,8 @@ pub fn scaffold_dna(
                 properties: None,
                 quantum_time: None,
             },
-            version: None,
+            installed_hash: None,
+            _version: None,
```
maackle, line 220: Shouldn't need it

```suggestion
```

robbiecarlton: Closing this PR as it's branched off of main, but I have a local branch with the same changes branched off of `holochain-0.1` and I'll your suggested change there.

PR #87, opened 2023-04-18
Fix transitive dependency not pinned

ThetaSinner:
`holochain_integrity_types` should be pinned but is not, this forced the transitive dependency onto a compatible version.

Temporary fix to get scaffolding working

### DIFF
--- a/src/scaffold/zome.rs
+++ b/src/scaffold/zome.rs
@@ -339,6 +339,12 @@ pub fn add_common_zome_dependencies_to_workspace_cargo(
         &"hdk".to_string(),
         &format!("={}", hdk_version()),
     )?;
+    // TODO temporary: force pin the version because HDI does not require a specific version (^0.1.1 currently)
+    let file_tree = add_workspace_external_dependency(
+        file_tree,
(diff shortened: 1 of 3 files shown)


guillemcordoba:
Why not just update the hdi to =0.2.2 ? I don't really want that crate as a dependency in all the apps...

ThetaSinner:
Understood, but the problem is actually in holochain because the `hdi` crate has been published with a `^` version  https://github.com/holochain/holochain/blob/develop-0.1/crates/hdi/Cargo.toml#L27 for `holochain_integrity_types` so unless we pin it somehow in the consumer it's going to be hard to fix.

For the version merged to `main` rather than this maintenance branch I don't think we should do this, I think we should publish a new version of the `hdi` which doesn't have this problem

guillemcordoba (approved):

ThetaSinner:
This is running into the `-lsqlite` issue that we're seeing elsewhere. Moving my focus to that

ThetaSinner:
Absorbed into #89 and #90, closing

PR #89, opened 2023-04-19
Holochain 0.1.4

robbiecarlton:
- Bumps scaffold and scaffolded happs to holochain 0.1.4
- Updates validation code to reflect this change https://github.com/holochain/holochain/pull/1910 in holochain
- (temporarily) Adds a pinned version of `holochain_integrity_types` to the scaffolded app, which can be removed once `hdi` has been updated to pin `holochain_integrity_types
- Adds extend-space action to fix disk space issue in CI


src/scaffold/dna.rs
```diff
@@ -217,6 +217,7 @@ pub fn scaffold_dna(
                 quantum_time: None,
             },
             installed_hash: None,
+            _version: None,
```
robbiecarlton, line 220: @maackle - tried removing this but the compiler complained

guillemcordoba (approved):

Cargo.toml
```diff
@@ -1,13 +1,14 @@
...
 name = "holochain_scaffolding_cli"
-version = "0.1.7"
+version = "0.1.8"
 description = "CLI to easily generate and modify holochain apps"
 license-file = "LICENSE_CAL-1.0"
 homepage = "https://developer.holochain.org"
 documentation = "https://docs.rs/holochain_scaffolding_cli"
 repository = "https://github.com/holochain/scaffolding"
 readme = "README.md"
+resolver = "2"
```
ThetaSinner, line 11: Technically unecessary because `edition = "2021"` uses this by default. You only have to set it in a workspace if you want this to apply from the top level. Won't do any harm though

flake.nix
```diff
@@ -25,7 +25,7 @@
           }: {
             devShells.default = pkgs.mkShell {
               inputsFrom = [ inputs.holochain.devShells.${system}.holonix ];
-              packages = [ pkgs.nodejs-18_x ];
+              packages = [ pkgs.nodejs-18_x pkgs.sqlite ];
```
ThetaSinner, line 28: Please could you add a note to take this out in favour of https://github.com/holochain/holochain/pull/2248 which I will backport asap?

ThetaSinner (approved):

zippy (approved):

PR #92, opened 2023-04-26
[Do not merge] Holochain 0.2

robbiecarlton:
This will be the release branch for scaffolding compatible with the latest `0.2.x` version of `holochain`, analogous with the `holochain-0.1` branch.

It's currently pointing to `0.2.0-beta-rc.6` so we can test against the rc.

This PR shouldn't be merged. I'm leaving it up for now for convenience of reading the diff.

Once we have an `0.2.0` release, this branch will use that, and development will continue on other branches.


robbiecarlton:
This is currently blocked, waiting for an update to hc launch

ThetaSinner:
I'm going to assume this is no longer relevant, there is now a `holochain-0.2` branch which points at a more recent version of 0.2.x and 0.2.0 never really worked anyway so even if this was still needed it should be updated to a newer version

PR #95, opened 2023-05-09
Fix bug in merge scope

guillemcordoba:


.github/workflows/test.yaml
```diff
@@ -30,5 +30,5 @@ jobs:
       - name: Install and test
         run: |
           cd $GITHUB_WORKSPACE
-          nix develop --command bash -c "cargo install --path . && sh run_test.sh"
+          nix develop --command bash -c "cargo t && cargo install --path . && sh run_test.sh"
```
steveeJ, line 33: what's `cargo t`?

flake.nix
```diff
@@ -25,7 +25,7 @@
           }: {
             devShells.default = pkgs.mkShell {
               inputsFrom = [ inputs.holochain.devShells.${system}.holonix ];
-              packages = [ pkgs.nodejs-18_x ];
+              packages = [ pkgs.nodejs-18_x pkgs.sqlite ];
```
steveeJ, line 28: this isn't required anymore since holochain was fixed to use bundled sqlite in all circumstnaces

.github/workflows/test.yaml
steveeJ: also please see https://github.com/holochain/scaffolding/issues/96

steveeJ (commented):
i haven't reviewd the rust changes so only commenting on the rest.

flake.nix
guillemcordoba: Yeah okey... In which version? Should we upgrade holochain in the main branch of the scaffolding tool?

steveeJ: here's a PR for that: #97

ThetaSinner:
@guillemcordoba What does this fix? So that I can give it a review and get it rebased/merged

guillemcordoba:
Yeah sorry. This fixes a bug in the `merge_scopes` handlebars helper, in which it would remove the next character from the opening scope string. So this:

```
const s = {};
```

When applied this template:
```
{{#merge_scope "const s = {" }}
};
{{/merge_scope}}
```

Would become

```
const s = {;
```

This PR fixes that bug.

ThetaSinner (approved):

ThetaSinner:
Thank you, understood :) I've rebased this and pointed it to the `develop` branch. It can be back-ported and released from there

PR #97, opened 2023-05-11
feat(flake,nix): bump to 0_2; feat(ci/test): use nix to build scaffolding

steveej:
* adapt to upstream flake changes
* bump the versions input to versions/0_2
* use nix to build scaffolding on CI, which can re-use cached dependencies accross runs

---

equivalent to #94 

- [ ] port scaffolding changes from `holochain-0.2`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,5 +30,5 @@ jobs:
       - name: Install and test
         run: |
           cd $GITHUB_WORKSPACE
-          nix develop --command bash -c "cargo install --path . && sh run_test.sh"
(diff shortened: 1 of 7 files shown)


guillemcordoba:
Mmm shouldn't this PR also bump the holochain version in scaffolding? Is the new way of running the test testing the code in this repository, or the binary coming from the holochain main repository?

steveeJ:
@guillemcordoba i did that and it leads to CI errors. given this implication, is the other way true as well? if we don't bump it but do bump the conductor and the template versions, will there be negative consequences for users?

ThetaSinner:
@steveej Do you know what from this PR is still relevant? I believe that the key aspects have made it into the repo on other PRs?

PR #100, opened 2023-05-31
feat(test): use scaffolding binary built by nix

steveej:
holonix comes with a version of scaffolding by default. overriding that has the same effect as building it manually from the repo and avoids potential double-compilation, _and_ allows cachix to cache the version.


guillemcordoba:
Oh so this builds and tests the code that exists in this repo?

steveeJ:
> Oh so this builds and tests the code that exists in this repo?

yes, the _holochain-0.2_ branch already has this change. i just realized it's not on _main_ either but it could be too.

ThetaSinner:
These changes are on the target branch, closing

PR #101, opened 2023-05-31
feat(test): use scaffolding binary built by nix

steveej:
holonix comes with a version of scaffolding by default. overriding that has the same effect as building it manually from the repo and avoids potential double-compilation, and allows cachix to cache the version.


ThetaSinner:
These changes are now on main, a rebase wipes them out

PR #116, opened 2023-09-15
Update CI and flake

ThetaSinner:


setup.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
steveej, line 1: accidental commit of this file?

ThetaSinner: Yes, was checking the Nix version, removed!

steveej:
what's missing here is a writable nix cache. at least the dependencies might not change on every PR.

ThetaSinner:
Definitely agree with that. It'd speed things up significantly. So would the suggestion to split the test script to run separate tests in parallel. I think this is enough for one PR and we should look at caching separately

steveej (approved):

robbiecarlton:
Closing, this work has been incorporated into holochain-0.1 and holochain-0.2 branches

PR #126, opened 2023-09-22
Prevent trailing commas when field list ends in non-widget fields in EditFoo.svelte template

pdaoust:
This is not the most elegant way to do things -- it duplicates data; implementing a `filter` helper for Handlebars would've been much more elegant. But the advantage of this approach is that I can actually do it ;)

I'm working on a possibly better solution that involves writing a `filter` Handlebars helper, but I'd like to get things unblocked for the hackathon this weekend

### DIFF
--- a/src/scaffold/entry_type.rs
+++ b/src/scaffold/entry_type.rs
@@ -101,6 +101,15 @@ pub fn scaffold_entry_type(
         }
     };
 
+    let widget_fields = fields
+        .iter()
+        .map(|f| f.clone())
+        .filter(|f| match f.widget {
+            Some(_) => true,
+            None => false
+        })
+        .collect();
+
     let reference_entry_hash = match maybe_reference_entry_hash {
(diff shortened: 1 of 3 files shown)


src/scaffold/entry_type.rs
```diff
@@ -101,6 +101,15 @@ pub fn scaffold_entry_type(
...
     };
 
+    let widget_fields = fields
+        .iter()
+        .map(|f| f.clone())
+        .filter(|f| match f.widget {
+            Some(_) => true,
+            None => false
+        })
+        .collect();
```
robbiecarlton, line 111:
```suggestion
    let widget_fields = fields.clone()
        .iter()
        .filter(|f| match f.widget {
            Some(_) => true,
            None => false
        })
        .collect();
```
Less code, and I'm not 100% but I think it will also be marginally more efficient

robbiecarlton (approved):
one non blocking suggestion

PR #128, opened 2023-09-28
Implement `filter` helper and use it to suppress reactive variables for non-widget fields

pdaoust:
#126 solved the problem of `EditFoo.svelte` templates creating trailing commas in the reactive variable declarations; e.g., for an entry type `Foo` with fields `visible_name` and `hidden_birthdate`, the template would generate this line:

```javascript
$: visible_name, ;
```

But it was an ugly solution. This introduces a better solution via a new `filter` helper for Handlebars. You pass it a value containing a list of items and a predicate, and it gives back a list of filtered items. Then you can filter out non-widget fields and use HBS' `last` helper to suppress the final comma. (I'm sure it has other uses too.)

It just renders a very tiny Handlebars template that uses `{{#if}}` to check the predicate against each item. This lets you use Handlebars' full functionality in the predicate.


pdaoust:
@robbiecarlton the biggest thing I'm worried about with this PR is that I kinda just cargo-culted the ownership/borrowing/consuming thing until `rustc` shut up. Not sure if what I did was good practice. And I also don't know about the iterator chain I used for filtering the items -- feels like there ought to be some more elegant way of filtering _and_ failing on the first error, but I tried a few ideas and then gave up. Currently it throws out runtime handlebars errors, but it does test the template once with the first item in the list to make sure the passed-in filter predicate actually works. (The way it works is that it just creates a tiny template, `{{#if <condition>}}true{{else}}false{{/if}}` and then passes each item to that template in turn.)

ThetaSinner:
@pdaoust A few things I'd like to do to get this merged

- Please can the PR target the `main` branch and I'll take care of merging it to release series branches
- Could you please give me some steps to reproduce the problem you're fixing so I can do my own manual testing?
- Are you up for adding some unit tests where appropriate for this? I'm not sure what's already here so it might not be super straight forward. I'm happy to either work on this with you or take it on myself based on some manual testing

ThetaSinner:
I'm going to go ahead and re-target this to `develop` because the current target branch is going to be renamed

ThetaSinner:
Okay, this is now ready for review and can be merged when ready

src/templates.rs
```diff
@@ -275,6 +276,83 @@ pub fn register_case_helpers<'a>(mut h: Handlebars<'a>) -> Handlebars<'a> {
...
+
+        let items: Vec<Value> = match value {
+            Value::Array(items) => Ok(
+                items
+                    .iter()
+                    .map(|item| item.clone())
+                    .collect()
+            ),
+            // FIXME: This doesn't preserve object keys.
+            // That's probably unexpected for consumers.
```
c12i, line 322: To fix this you could perhaps consider creating a `FilterableValue` enum with the following variants:

```rs
pub enum FilterableValues {
    Array(Vec<Value>),
    Object(Map<String, Value>)
}
```

Then instead of assigning items as a `Vec<Values>` you replace that with the new enum:

```rs
let items: FilterableValues = match value {
    Value::Array(items) => Ok(FilterableValues::Array(items.iter().cloned().collect())),
    Value::Object(items) => Ok(FilterableValues::Object(items.clone())),
     _ => Err(
         RenderError::new("Filter helper: value to be filtered must be an array or object")
     ),
  }?;
```

then each variant can be handled like so:

```rs
match items {
    FilterableValues::Array(items) => {
        let mut filtered_array = vec![];
        for item in items.iter() {
            match r.render_template(&template, item) {
                Ok(s) => {
                    if s.as_str() == "true" {
                        filtered_array.push(item)
                    }
                }
                Err(e) => {
                    return Err(e);
                }
            }
        }
        Ok(ScopedJson::Derived(json!(filtered_array)))
    },
    FilterableValues::Object(object) => {
        let mut filtered_object = Map::new();
        for key in object.keys() {
            if let Some(v) = object.get(key) {
                if v.as_str().ok_or(RenderError::new("something"))? == "true" {
                    filtered_object.insert(key.into(), v.clone());
                }
            }
        }
        Ok(ScopedJson::Derived(json!(filtered_object)))
    }
}
```

Note that I have not really tested this code, so take this more as a recommendation, tell me what you think

src/templates.rs
```diff
@@ -275,6 +276,83 @@ pub fn register_case_helpers<'a>(mut h: Handlebars<'a>) -> Handlebars<'a> {
...
+        let value = value_param.value();
+
+        let condition_param = h
+            .param(1)
+            .ok_or(RenderError::new("Filter helper: Param not found for index 1; must be string containing filter condition predicate"))?;
+
+        let condition = condition_param
+            .value()
+            .as_str()
+            .ok_or(RenderError::new("Filter helper: filter condition predicate must be a string"))?;
```
c12i, line 307: This is more of an aesthetic suggestion to keep code clean and prefer iterables

```suggestion
        let mut params = h.params().iter();
        let value = params
            .next()
            .ok_or(RenderError::new(
                "Filter helper: Param not found for index 0; must be value to be filtered",
            ))?
            .value();

        let condition = params
            .next()
            .ok_or(RenderError::new("Filter helper: Param not found for index 1; must be string containing filter condition predicate"))?
            .value()
            .as_str()
            .ok_or(RenderError::new("Filter helper: filter condition predicate must be a string"))?;
```

c12i (changes requested):
Better solution than what is proposed in #130 since we skip adding a `widget_fields` field to the `EntryDefinition` struct. I lean more towards this option as well

However, @ThetaSinner is right, some unit tests for the `filter` helper will definitely be needed, with cases covering usage of the `includeZero` optional parameter

@pdaoust kindly let me know if you are able to resolve the merge conflicts in this PR too

src/templates.rs
```diff
@@ -275,6 +276,83 @@ pub fn register_case_helpers<'a>(mut h: Handlebars<'a>) -> Handlebars<'a> {
...
+
+        let filtered_items: Vec<&Value> = items
+            .iter()
+            .filter(|item| match r.render_template(&*template, item) {
+                Ok(s) => s.as_str() == "true",
+                // FIXME: this swallows error messages that don't involve parsing the template.
+                _ => false
+            })
+            .collect();
+        return Ok(ScopedJson::Derived(json!(filtered_items)));
```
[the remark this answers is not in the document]
pdaoust: Thanks; I struggled over this for a long time before giving up. Guess I don't need to be scared of ever using a `mut` :)

src/templates.rs
pdaoust: This is both a review and an education; thank you for all this!

pdaoust: I've implemented the suggestions and it compiles; I haven't tested it yet though (still trying to remember how I tested it in the past!)

c12i: You can test it out in with by compiling a handlebars template as a string, similar to how it was done here

https://github.com/holochain/scaffolding/blob/develop-0.2/src/templates/helpers/merge.rs#L279

pdaoust:
Okay, I've tidied up all the code per your feedback @c12i and I've also merged in `develop` -- I noticed that the helpers code was rearranged, so I moved the new `filter` helper into a similar structure (it gets its own file, figure it was sufficiently complex).

I'm pretty new to writing Rust, but I think I could come up with some tests that at the very least test the initial problem that this thing was trying to fix. How far do you think I should go -- wondering if it makes sense to test all possible falsey HBS values and a few representative truthy values, or if that's just overkill.

c12i:
> I noticed that the helpers code was rearranged, so I moved the new filter helper into a similar structure (it gets its own file, figure it was sufficiently complex).

Ah yes, should have mentioned this earlier, glad you made the change

> How far do you think I should go -- wondering if it makes sense to test all possible falsey HBS values and a few representative truthy values, or if that's just overkill.

No worries @pdaoust, we can test the happy path cases for now

src/templates/helpers/filter.rs
```diff
@@ -0,0 +1,114 @@
...
+    ScopedJson,
+};
+use serde_json::{json, Value, Map};
+
+#[derive(Clone, Copy)]
+pub struct FilterHelper;
+
+pub enum FilterableValues {
+    Array(Vec<Value>),
+    Object(Map<String, Value>), // imported from serde_json
```
c12i, line 12: Feel free to remove this comment

```suggestion
    Object(Map<String, Value>),
```

c12i (changes requested):

c12i (approved):

pdaoust:
@c12i could you review again? I simplified some code -- realised that a couple suggestions above tackled one problem in two different ways and cancelled each other out :) I also wrote a few basic tests and discovered/fixed a bug.

c12i (approved):

c12i:
> realised that a couple suggestions above tackled one problem in two different ways and cancelled each other out 

thanks for the changes @pdaoust, the implementation is much simpler now

PR #129, opened 2023-09-29
port `filter` helper to `holochain-0.2` branch

pdaoust:
Companion to #128. I haven't smoke-tested it, but it does pass local CI.


ThetaSinner:
@pdaoust Thank you for creating this. I would like to get the changes onto `main` first then we'll port the commits from there. That should make it easier to keep track of for maintenance than having two separate changes targeting different branches

PR #130, opened 2023-10-03
Template sweep -- fix bugs in Vue and Lit

pdaoust:
I don't claim this is a thorough sweep that catches all bugs, but it does catch all bugs discovered via following the Getting Started Guide but choosing Lit and Vue instead of Svelte for the templating engine.


ThetaSinner:
@pdaoust Same with this one, please can it target `main`? Then would you be willing to go through and comment on the PR with a description of what is being changed/fixed so I can do some testing?

I'm happy reviewing the code but I don't know the repo well enough to understand what's changing without having a conversation about it if that's okay! :)

ThetaSinner:
This is now ready for review and can be merged when ready

src/scaffold/entry_type.rs
```diff
@@ -101,6 +101,15 @@ pub fn scaffold_entry_type(
...
     };
 
+    let widget_fields = fields
+        .iter()
+        .map(|f| f.clone())
+        .filter(|f| match f.widget {
+            Some(_) => true,
+            None => false
+        })
+        .collect();
```
c12i, line 111: A little more cleaner 🙂

```suggestion
    let widget_fields = fields
        .iter()
        .cloned()
        .filter(|f| f.widget.is_some())
        .collect();
```

c12i (changes requested):
Discovered similar bugs, at least for the `Vue` and `Svelte` templates hence why I am returning to this PR, I think we should be getting this PR merged as it fixes some critical bugs

@pdaoust kindly let me know if you are in a position to resolve the merge conflicts

src/scaffold/entry_type.rs
c12i: Taking back this comment, perhaps we won't need this fix if the changes #128 are approved

c12i:
@pdaoust, allow me to close this PR since it is stale at the moment and most of the changes made here have long been addressed

pdaoust:
Sure @c12i . Are there any fixes that haven't already been made? Should I do a new PR with those?

c12i:
Thanks @pdaoust, I don't think there are any changes at the moment, I inadvertently addressed them while transitioning the templates to the new styling.

PR #131, opened 2023-10-03
port #130 to Holochain 0.2

pdaoust:
Does what it says on the tin. #130


ThetaSinner:
Closing to allow back-porting to happen later. Let's get #130 merged first

PR #134, opened 2023-10-06
Feat/replace test pause with dhtsync

mattyg:
Repeat of #105 where I borked the git history @robbiecarlton

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1704,17 +1704,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "errno"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50d6a0976c999d473fe89ad888d5a284e55366d9dc9038b1ba2aa15128c4afa0"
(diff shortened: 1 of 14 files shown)


ThetaSinner (approved):

ThetaSinner:
Please can this target `main` instead of `develop`? This repo needs a contribution guide but I'm not sure the branches are being used consistently even by maintainers yet. I had a conversation about this a while ago and I'll check on that and get it into a short contribution guide.

ThetaSinner:
I tried changing the base but it makes a mess to just do that because `develop` is out of sync with `main`

ThetaSinner:
I'm just discovering that develop and main appear to be out of sync. I assume one was the upstream of the other. I'm going to get them tidied up so hopefully this becomes a simple base change

ThetaSinner:
I've replaced this PR with these same commits and synced with the updated base branch. Normally wouldn't do that but I'm trying to get develop synced up so I can proceed with tweaking the branching strategy on the repo :)

PR #137, opened 2023-10-25
Update vue scaffolding

emhoracek:
I ran into some trouble trying out the tutorial forum example using the Vue scaffolding instead of svelte, so I tried to fix it. I couldn't figure out a way to easily try out my fixes without making an option to not make a nix environment, which is why I added the setup_nix option, but I can revert that if it's not useful. I think I got all the places that using `error.data.data`, but there are other places that have the extra `Hash` (making `postHashHash` instead of `postHash`).

### DIFF
--- a/templates/vue/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{pascal_case collection_name}}.vue.hbs
+++ b/templates/vue/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{pascal_case collection_name}}.vue.hbs
@@ -4,7 +4,7 @@
   </div>
 
   <div v-else style="display: flex; flex-direction: column">
-    <span v-if="error">Error fetching the {{lower_case (plural referenceable.name)}}: {{{{raw}}}} {{error.data.data}}.{{{{/raw}}}}</span>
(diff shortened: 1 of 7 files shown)


pdaoust:
Thanks for submitting this @emhoracek ! I discovered the same thing when I was doing a QA sweep of the scaffold templates (using the same technique you did -- following the tutorial with another front-end framework). The fact that you discovered and fixed the same issues suggests that my PR didn't get merged... hm

Looking at your PR, it also looks like you discovered some errors in error display cases (ironic) that I never hit. So I'd like to merge your work here...

Almost all of it, that is! looks like you've also added the option to scaffold holonix env for the example hApps, which is valuable on its own. Could you split that into a separate PR?

emhoracek:
> Almost all of it, that is! looks like you've also added the option to scaffold holonix env for the example hApps, which is valuable on its own. Could you split that into a separate PR?

Sure, just made #138 

I'm going to go again and fix the other places that have that `HashHash` bug as well.

ThetaSinner:
It took me a bit of time to figure out what I thought was a bug with scaffolding while testing this. Turns out it was user error on my part :) Giving this a look now

ThetaSinner (approved):

ThetaSinner:
Looks great, thank you!

PR #138, opened 2023-10-29
Setting up nix is optional for example apps

emhoracek:
I just did this to make it easier to test out the fixes I made to the Vue scaffolding. I'm new to Nix so I'm not sure it was necessary - maybe there's some other work-around. The problem was that I made a Nix environment so I could build the project, but then I couldn't run my compiled code because it was trying to start a Nix environment inside the Nix environment. So I added this flag so I could tell the example set-up not to setup Nix.


ThetaSinner:
One small change in two places please. Where the `scaffold_web_app` function is called, it's currently hard-coded to pass false for the `skip_nix` parameter. Please could that be replaced with something like `!setup_nix.unwrap_or(true)`.

With that change I think this is good to go :)

c12i (changes requested):

c12i:
@emhoracek any chance you could give this PR a quick update and tackle the feedback? Happy to get this merged soon.

c12i (approved):

PR #143, opened 2023-11-06
Tidy up license file by referencing the published one

ThetaSinner:


ThetaSinner:
Also did a flake update because the project isn't building with an older Rust version so I need that

PR #144, opened 2023-11-06
Bump next version

ThetaSinner:
The versions between Holochain 0.1 and Holochain 0.2 aren't distinguished but it is important which version of scaffolding works against each. I propose we follow Holochain's minor version which makes it easy to match them. The patch versions need not match and can be incremented with each scaffolding release for that version of Holochain

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 edition = "2021"
 name = "holochain_scaffolding_cli"
-version = "0.1.11"
+version = "0.2.0"
 description = "CLI to easily generate and modify holochain apps"
 license-file = "LICENSE_CAL-1.0"
 homepage = "https://developer.holochain.org"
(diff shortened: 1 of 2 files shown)


ThetaSinner:
Also slipping in a flake update here so that the build will work

abe-njama:
@ThetaSinner thanks for raising this draft PR. Norbu says that he still needs to do some thorough testing of the newer versions of Holochain before we can proceed to merge this PR. He should be getting to testing within the next week.

ThetaSinner:
@abe-njama merging this wouldn't release a new version. The point of this PR is to address the versioning strategy being used for scaffolding which is really separate from any testing that needs to be done for the code.

There is an open discussion in MM in the scaffolding channel to be resolved.

ThetaSinner:
Not the versioning strategy that has been selected, closing

PR #145, opened 2023-11-06
Add initial maintenance guide

ThetaSinner:
I think this covers most things for a first pass. One notable thing missing is what 'bump the scaffolding version' means. This is an open discussion so I'll leave it out of the guide for the moment.

### DIFF
--- a/guides/maintenance.md
+++ b/guides/maintenance.md
@@ -39,7 +39,7 @@ Moving a release tag example:
 git checkout develop-0.1
 git pull --tags
 git tag holochain-0.1
-git push --tags origin holochain-0.1
+git push -f --tags origin holochain-0.1
 ```
 
 There is a CI job for Holonix to periodically (every 6 hours at the time of writing) looks for changes. It will automatically update its inputs for Scaffolding and create+merge a PR on the Holochain repository. So after you have pushed the updated tag, Holonix users will have to wait for the next automatic update then do their own `nix flake update` to see the latest Scaffolding.
--- a/guides/maintenance.md
+++ b/guides/maintenance.md
@@ -39,7 +39,7 @@ Moving a release tag example:
 git checkout develop-0.1
 git pull --tags
 git tag holochain-0.1
-git push -f --tags origin holochain-0.1
+git push --force --tags origin holochain-0.1
 ```
 
 There is a CI job for Holonix to periodically (every 6 hours at the time of writing) looks for changes. It will automatically update its inputs for Scaffolding and create+merge a PR on the Holochain repository. So after you have pushed the updated tag, Holonix users will have to wait for the next automatic update then do their own `nix flake update` to see the latest Scaffolding.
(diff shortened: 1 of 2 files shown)


ThetaSinner:
I'd actually quite like a document like this for Holochain. I think we have pieces of it but I'm not sure we've actually written down our PR/back-port process publicly. There are some HackMD docs and an agreement among the team but it'd be good to get it written down

guides/maintenance.md
```diff
@@ -0,0 +1,45 @@
...
+2. Perform any testing which needs to be done before releasing the new changes.
+2. Changes on a maintenance branch such as `develop-0.1` are not yet visible to Holonix because there is a tag on the branch with the same version number such as `holochain-0.1`. This is the marker that Holonix will use to fetch the current version of Scaffolding corresponding to its Holochain version. Move this label to the tip of the maintenance branch. Sample commands are given for this below.
+
+Moving a release tag example:
+
+```bash
+git checkout develop-0.1
+git pull --tags
+git tag holochain-0.1
+git push --tags origin holochain-0.1
```
steveej, line 42: IIUC this requires `--force` to overwrite the tag.

steveej (commented):
i want to read over this again but spotted something that i'd like to report right away

guides/maintenance.md
ThetaSinner: Good spot thank you. I suspect this guide will need some pipe cleaning and updates as we try to use it but would prefer to catch as much of this kind of thing as possible up front :)

guides/maintenance.md
```diff
@@ -0,0 +1,45 @@
...
+2. Perform any testing which needs to be done before releasing the new changes.
+2. Changes on a maintenance branch such as `develop-0.1` are not yet visible to Holonix because there is a tag on the branch with the same version number such as `holochain-0.1`. This is the marker that Holonix will use to fetch the current version of Scaffolding corresponding to its Holochain version. Move this label to the tip of the maintenance branch. Sample commands are given for this below.
+
+Moving a release tag example:
+
+```bash
+git checkout develop-0.1
+git pull --tags
+git tag holochain-0.1
+git push --force --tags origin holochain-0.1
```
steveej, line 42: i'm pretty sure `--tags` isn't required and therefore i'd prefer omitting it to not accidentally push other local tags that might not be desired on the remote.

```suggestion
git push --force origin holochain-0.1
```

guides/maintenance.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Branching strategy
+
+New development and bug fixes should target the `develop` branch. From there, changes are back-ported to maintanance branches which are named after the version of Holochain they work with. So maintance for the versions of Scaffolding that work with Holochain 0.1.x are done on `develop-0.1` and those that work with Holochain 0.2.x are on `develop-0.2` and so on.
+
+There need to be some exceptions to this workflow to allow changes on maintenance branches that don't make sense to make on `develop`:
+
+- Bug fixes that are specific to a particular version of Holochain. This should be rare, but if Holochain were to introduce an API change between 0.1.x and 0.1.x+1 but NOT make that change for Holochain 0.2.x then the PR for Scaffolding would target `develop-0.1` directly.
+- Version bumps for `flake.nix`, `flake.lock` and `Cargo.lock` files should target the maintenance branches directly.
+- Version bumps for dependencies of Scaffolding in the `Cargo.toml`. For dependencies that still on the same version on `develop` then it would be valid to target `develop` and back-port but you'll end up with a tricky lock file merge anyway.
```
steveej, line 13:
```suggestion
- Version bumps for dependencies of Scaffolding in the `Cargo.toml`. For dependencies that are still on the same version as on `develop` it would be valid to target `develop` and back-port but you'll end up with a tricky lock file merge anyway.
```

steveej (approved):
looks to be consistent including explaining expected exceptional cases to the default too, thanks! :+1: i left some small remarks as suggestions

ThetaSinner:
Thanks for the review, I've taken the suggestions and will go ahead with getting this in :)

PR #154, opened 2023-11-15
Refactor gets

guillemcordoba:
Summary of changes:

- Updated `syn` and `pretty-please` dependencies, this is needed to handle the new rust `let Some(_) = my_option else {};` statements.
- Disambiguated the `get_entry_type_name` getters:
  - Previously, we only had a `get_entry_type_name(action_hash: ActionHash) -> ExternResult<Option<Record>>` that:
    - If the entry was deleted, returned None
    - Otherwise, would return the latest update for the entry
  - This is bad for a couple of reasons:
    - Whenever there is an update to the entry, there is actually no way for us to get the original action, so things like the creation timestamp are not accessible anymore, there are no calls that retrieve it
    - Once the entry is deleted, there is no way to see it again, if you want to implement an "archived" list of entries 
    - There is no `get_all_updates` getter, so the UI cannot show a list of updates
  - So this PR disambiguates that single get into multiple ones:
    - If the entry can be updated:
      - `get_latest_entry_type_name(action_hash: ActionHash) -> ExternResult<Option<Record>>`: gets the latest update for the entry, returns the entry even if it's deleted
      - `get_all_entry_type_name_revisions(action_hash: ActionHash) -> ExternResult<Vec<Record>>`: gets the list of updates for the entry, returns them even if it's deleted
      - `get_original_entry_type_name(action_hash: ActionHash) -> ExternResult<Option<Record>>`: gets the create record for the entry, returns the entry even if it's deleted
    - If the entry cannot be updated:
      - `get_entry_type_name(action_hash: ActionHash) -> ExternResult<Option<Record>>`: gets the create record for the entry, returns the entry even if it's deleted
    - If the entry can be deleted:
      - `get_deletes_for_entry_type(action_hash: ActionHash) -> ExternResult<Vec<SignedActionHash>>`: returns the array of `Delete` actions for the entry, allowing the UI to display information like "it was deleted 4 minutes ago", that was previously not accessible

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -420,7 +420,7 @@ checksum = "a66537f1bb974b254c98ed142ff995236e81b9d0fe4db0575f46612cb15eb0f9"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.38",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -880,7 +880,7 @@ dependencies = [
  "heck 0.4.1",
  "proc-macro2",
  "quote",
- "syn 2.0.38",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -1352,7 +1352,7 @@ dependencies = [
  "ident_case",
  "proc-macro2",
  "quote",
- "syn 2.0.38",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -1396,7 +1396,7 @@ checksum = "836a9bbc7ad63342d6d6e7b815ccab164bc77a2d95d84bc3117a8c0d5c98e2d5"
 dependencies = [
  "darling_core 0.20.3",
  "quote",
- "syn 2.0.38",
+ "syn 2.0.39",
 ]
 
 [[package]]
@@ -1468,7 +1468,7 @@ checksum = "67e77553c4162a157adbf834ebae5b415acbecbeafc7a74b0e886657506a7611"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.38",
+ "syn 2.0.39",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


mattyg:
I see where you're coming from. I wonder if there should also be a `get_earliest_delete_for_entry_type(action_hash: ActionHash)` since mirroring the function that gives an opinionated "latest" update.

Anyway since this doesn't actually remove the existing function and just renames it + adds additional ones, I'm fine with it. I'd have to play around with it to get a feel for it anyway.

guillemcordoba:
Mmm interesting... Did you find yourself needing that function at some point?

mattyg:
> Mmm interesting... Did you find yourself needing that function at some point?

Yeah I just think the standard use case is only caring about the oldest delete because that's what you would treat as the "canonical" timestamp and delete author. Having multiple deletes that are all meaningful is an edge use case.

It also feels weird to me that you're including an opinionated way to define the "latest" update, while not also providing an opinionated way to define the "latest" delete -- either have both or neither.

guillemcordoba:
@ThetaSinner @mattyg I'd like to get this and a couple other PRs merged as other work I'm doing depends on them. Could I get a review / go ahead?

guillemcordoba:
Yeap! Your proposal makes perfect sense @mattyg , and it aligns with the way that updates work. So I'll make that change and then we can merge this. Thanks for your feedback.

guillemcordoba:
@ThetaSinner this is ready to be merged AFAICT. Can I get get a review? Also, what's the policy on updating the holochain-0.x branches? I only need this to be included in the holochain-0.2 one, and maybe we don't want to backport it to 0.1 since it'd be a breaking change to people with already scaffolded apps?

ThetaSinner (approved):
Looks great!

guillemcordoba:
Great, can I merge it then @ThetaSinner ?

guillemcordoba:
Amazing, thanks! One less to go.

PR #158, opened 2023-11-21
Fix Vue bool and timestamp

ThetaSinner:
Ran into these while testing another change. It looks like these types aren't covered by the automated tests, something that would be good to cover because these typescript errors prevent the app running


templates/vue/field-types/Timestamp/DateTimePicker/detail/render.hbs
```diff
@@ -1 +1 @@
-{{{{raw}}}} {{ {{{{/raw}}}} new Date({{variable_to_read}} / 1000).toLocaleString() {{{{raw}}}} }} {{{{/raw}}}}
+{{{{raw}}}} {{ {{{{/raw}}}} {{variable_to_read}} ? new Date({{variable_to_read}} / 1000).toLocaleString() : 'Missing timestamp' {{{{raw}}}} }} {{{{/raw}}}}
```
ThetaSinner, line 1: Produces 

```
<span style="white-space: pre-line">{{  jklh?.when ? new Date(jklh?.when / 1000).toLocaleString() : 'Missing timestamp' }} </span>
```

Instead of 

```
<span style="white-space: pre-line">{{  new Date(jklh?.when / 1000).toLocaleString() }} </span>
```

Which is not valid TypeScript as you can't divide undefined by a number

guillemcordoba (approved):
Awesome!

PR #159, opened 2023-11-21
Start a manual testing guide

ThetaSinner:


guillemcordoba (approved):

guillemcordoba:
I see no problem with this, although I've never needed to do anything like this, as the tests are broad and test all the features.

ThetaSinner:
That's exactly it, this is when I want to run the latest scaffolding and take a look at the app that's produced then this is useful. I prefer to rely on automated testing too but I think it's still useful to be able to do this :)

PR #160, opened 2023-11-28
Nov 0.1 backports

ThetaSinner:
All outstanding back-ports to 0.1 as of today


ThetaSinner:
Can't backport the pause/dhtSync change because tryorama 0.11.x doesn't have `dhtSync`

PR #162, opened 2023-12-06
Adds holo-dev-server into the nix develop shell if scaffold is holo-enabled

alastairong:
For holo-enabled scaffolds we need to make holo-dev-server available to the developer for app testing.

This PR makes it available via the nix shell.


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 edition = "2021"
 name = "holochain_scaffolding_cli"
-version = "0.1.11"
```
ThetaSinner, line 4: We shouldn't make version changes on this branch please, the versions need to be maintained separately on the release branches and changing it here risks mistakes in the back-port

src/scaffold/app/nix.rs
```diff
@@ -9,47 +9,76 @@ use crate::error::{ScaffoldError, ScaffoldResult};
...
+    };
+
+    let extra_content = if holo_enabled {
+        r#"extraSubstitutors = [ "https://cache.holo.host" ];
+                trustedPublicKeys = [
+                    "cache.holo.host-2:ZJCkX3AUYZ8soxTLfTb60g+F3MkWD7hkH9y8CgqwhDQ="
+                ];
+
+                shellHook = ''
+                    nix-env -f "https://hydra.holo.host/channel/custom/holo-nixpkgs/alpha/holo-nixpkgs/nixexprs.tar.xz" -iA holo-dev-server
```
ThetaSinner, line 29: I think this is potentially surprising for updated versions of a binary to be installed on starting a shell. Is there another way of doing this so it's included as a package and locked with the flake?

ThetaSinner (commented):
Please could the tests be extended to cover this new code?

Cargo.toml
alastairong: Ok will revert

src/scaffold/app/nix.rs
alastairong: Should it be locked with the flake? This binary should represent production Holo Network, and that can change.

alastairong: That said, adding it in this way seems ugly to me. I just don't know a better way

ThetaSinner: After discussing I understand why this is hard to change. With the source code being closed source it's not simple to build from source and hydra is giving us only the latest binary per channel so we don't have an obvious way to lock to a specific version. It would be good to solve this at some point so that people can look at the flake.lock and see what version they have, and when it gets updated. This also makes offline development difficult but for now the install line could be commented out temporarily and the binary manually installed before going offline or some other workaround.

steveej: this alters the user's profile permanently and not just for the duration of the nix shell. what is the chance to open-source the holo-dev-server?

alastairong: eek

steveej: packaging closed-source binaries isn't great but it's feasible. it does require some form locking, which can be automatically maintained on a separate public repo and consumed here. given that the scaffolding performs a `nix flake update` makes the consumption of it a non-issue.

ThetaSinner:
The test looks good to me, and the bash logic to check if the program is available looks sound to me. Not sure why it's not passing though?

ThetaSinner (approved):

alastairong:
@steveej Hopefully we can remove that `builtIns.currentSystem` from the holochain flake and then no longer need to add --impure. Otherwise I'll need to update documentation accordingly.

c12i:
@alastairong I think the changes in #165 can also be included in this PR, seeing they are very similar

alastairong:
Deprecating in favour of #181 as Numtide is adding their commits there.

PR #170, opened 2024-01-02
Remove duplicate unit tests and run them in ci

c12i:
This PR removes duplicate unit tests in `templates.rs` and refactors existing unit tests. Also, the unit tests are now run in a new `cargotest` workflow.


.github/workflows/test.yaml
```diff
@@ -9,6 +9,22 @@ on:
     branches: [ develop, develop-0.1, develop-0.2 ]
 
 jobs:
+  cargotest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions-rs/toolchain@v1
```
ThetaSinner, line 17: Both of these `actions-rs/*` actions have had their repositories archived late last year. Probably fine for now but it would be worth thinking about alternatives in the near future

src/templates.rs
```diff
@@ -295,89 +295,3 @@ pub fn choose_or_get_template(
...
-
-        let (scope_opener_index, scope_close_index) =
-            get_scope_open_and_close_char_indexes(&text, &scope_opener).unwrap();
-
-        assert_eq!(scope_opener_index, 10);
-        assert_eq!(scope_close_index, 11);
-    }
-
-    #[test]
-    fn test_merge_match_scope() {
```
ThetaSinner, line 319: I agree the test above is duplicated but this one was added recently and looks like it has differences to the other `test_merge_match_scope`. It would be worth checking with @guillemcordoba on this test removal I think

ThetaSinner: It looks plausible that the differences are cosmetic and the tests are actually checking the same thing but I believe this was added to close a bug recently so I'd like to check

guillemcordoba (approved):
@c12i is right, those tests are duplicated. In [this PR](https://github.com/holochain/scaffolding/pull/157) I moved them [here](https://github.com/holochain/scaffolding/blob/4b14a8e1728bddf145baa5534165f2b90120d1b1/src/templates/helpers/merge.rs#L236), so this PR looks fine to me.

.github/workflows/test.yaml
```diff
@@ -9,6 +9,22 @@ on:
     branches: [ develop, develop-0.1, develop-0.2 ]
 
 jobs:
+  cargotest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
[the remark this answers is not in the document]
c12i: Updated

.github/workflows/test.yaml
c12i: Thanks for the heads up, was not aware of this, better to use maintained actions, updated

ThetaSinner (approved):
🎆

PR #171, opened 2024-01-03
Handle error creating linked `ActionHash` or `EntryHash` without previous entry-types existing

### LINKED ISSUE #136: Trying to scaffold an entry-type with an `ActionHash` field that gets linked from its entry without any previous entry-types in the zome fails
This is the error: 

```
[holonix:~/projects/holochain/open-dev/test/cancellations]$ hc scaffold entry-type
✔ Entry type name (snake_case): · cancellation
✔ Multiple integrity zomes were found in DNA cancellations_test, choose one: · cancellations_integrity

Which fields should the entry contain?

✔ Choose field type: · String
✔ Field name: · reason
✔ Should this field be visible in the UI? · yes
✔ Choose widget to render this field: · TextArea

✔ Choose field type: · ActionHash
✔ Should a link from the ActionHash provided in this field also be created when entries of this type are created? · yes
Error: IO error: Empty list of items given to `Select`

Caused by:
    Empty list of items given to `Select`
```

Steps to reproduce:

```
nix run github:/holochain/holochain#hc-scaffold -- web-app testest
cd testest
nix develop
hc scaffold dna testest
hc scaffold zome testest
hc scaffold entry-type testest
```

And in that last command select a field of type `ActionHash` and say yes to the question "Should a link from the ActionHash provided in this field also be created when entries of this type are created?".

c12i:
This PR fixes #136

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -559,7 +559,7 @@ Add new entry definitions to your zome with:
                 let file_tree = load_directory_into_memory(&current_dir)?;
                 let template_file_tree = choose_or_get_template_file_tree(&file_tree, &template)?;
 
-                let name: String = match name {
+                let name = match name {
                     Some(n) => {
                         check_case(&n, "entry type name", Case::Snake)?;
                         n
--- a/src/file_tree.rs
+++ b/src/file_tree.rs
@@ -119,7 +119,7 @@ pub fn insert_file_tree_in_dir(
 }
 
 pub fn find_files_by_name(file_tree: &FileTree, file_name: &PathBuf) -> BTreeMap<PathBuf, String> {
-    find_files(file_tree, &|file_path, _file_contents| {
+    find_files(file_tree, &|file_path, _| {
         file_name.file_name().eq(&file_path.file_name())
     })
 }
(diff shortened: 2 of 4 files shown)


ThetaSinner:
This nicely fixes the error messaging when there are no other entry types defined yet. Is the entry type that is currently being defined not a valid target though? I'd expect it to be possible to create referential links between entries of the same type. So I'd actually expect this to always return at least the entry type that is currently being defined rather than an error

ThetaSinner:
I'd be happy to approve this as a fix for the error messages, but I wonder whether the issue should stay open to address the missing functionality if it is indeed desirable

ThetaSinner (approved):

guillemcordoba:
It's actually not valid @ThetaSinner, in the case of `ActionHash`. It _is_ valid when it's `Option<ActionHash>` or `Vec<ActionHash>` though. This is because if an entry depended always on an entry of its same type, that's actually self-referential infinite regression - you can't ever create the first entry of that type. @c12i are you taking that into account in this PR?

c12i:
> It is valid when it's `Option<ActionHash>` or `Vec<ActionHash>`

Interesting, I had not taken this into account, does this also apply for the `EntryHash` type, since these cases are handled together?

c12i:
@guillemcordoba the check has now been moved to account for self referencing if the selected type is an `Option<ActionHash>` or `Vec<ActionHash>`

PR #177, opened 2024-01-11
Backport `0.1`

c12i:


templates/vanilla/example/ui/index.html.hbs
```diff
@@ -45,8 +45,8 @@
       }
       
       async function connect() {
-        // We pass '' as url because it will dynamically be replaced in launcher environments
-        client = await AppAgentWebsocket.connect('', 'hello-world');
+        // We pass an unused string as the url because it will dynamically be replaced in launcher environments
+        client = await AppAgentWebsocket.connect(new URL('https://UNUSED'), 'hello-world');
```
ThetaSinner, line 49: That's curious, this wasn't broken on develop-0.1 then because it was the extract `const client =` that was causing the problem.

I don't think you want to make this a URL here - the 0.1 compatible client expects a string https://github.com/holochain/holochain-client-js/blob/v0.12.5/src/api/client.ts#L112C9-L112C9

ThetaSinner (changes requested):
Check compatibility with the js client for 0.1 please, otherwise looks good

c12i:
> Would it be possible to back-port this one? https://github.com/holochain/scaffolding/pull/141

Sure I guess this was unintentionally omitted in #160?

templates/vanilla/example/ui/index.html.hbs
c12i: Good catch, I guess the PR that introduced this change was mislabled

reverted

ThetaSinner: Certainly unnecessary on this branch which I didn't realise yes, but this wasn't part of that change, just a difference between branches https://github.com/holochain/scaffolding/pull/168/files#diff-547e152c3198bfe30e49e633ef3090b4a610bd1e180adb53e4074fa9b42be0aeR49

ThetaSinner:
> Sure I guess this was unintentionally omitted in https://github.com/holochain/scaffolding/pull/160?

Oh good spot, there is a comment on #160 about why that wasn't backported, sorry it also won't work with the older client. I think I left that label in place with the intention to backport the `dhtSync` change... But given how close we are to releasing 0.3 I wonder if that's really worth it now

ThetaSinner (approved):
Looks good!

PR #178, opened 2024-01-11
Backport `0.2`

c12i:


ThetaSinner (approved):
Interesting to see that the hello world fix wasn't needed on `develop-0.2`, it was only broken on develop. Feeling like my concerns about branch and change management were justified!

This backport looks great!

PR #179, opened 2024-01-12
Update maintenance docs

c12i:


guides/maintenance.md
```diff
@@ -38,7 +38,7 @@ Moving a release tag example:
 ```bash
 git checkout develop-0.1
 git pull --tags
-git tag holochain-0.1
+git tag holochain-0.1 --force
```
ThetaSinner, line 41:
```suggestion
git tag --force holochain-0.1
```

ThetaSinner (approved):
Proposed ordering flags and args the same way as the line below for consistency but up to you. Looks good thanks!

PR #165, opened 2023-12-19
--holo flag includes holo-dev-server in flake.nix

alastairong:

PR #181, opened 2024-01-15
Add holo-dev-server to the dev shell when --holo is passed

r-vdp:
We add a --holo flag to the example subcommand.
When this flag is passed, we add holo-nixpkgs as a flake input and we add holo-dev-server to the dev shell's packages.

This PR builds on #165 that was opened by @alastairong

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -83,8 +83,56 @@ setup_and_build_hello_world() {
   cd ..
 }
 
-if [[ -n "$SCOPE" && "$SCOPE" == "hello_world" ]]; then
-  setup_and_build_hello_world
+if [[ -n "$SCOPE" ]]; then
+
+  case "$SCOPE" in
+  "hello_world")
+    setup_and_build_hello_world
+    ;;
+  "holo_integration")
+    rm -rf /tmp/holo-flake
+    cd /tmp
+
+    hc-scaffold --template vue web-app holo-flake --setup-nix true
+    cd holo-flake
+
+    nix develop --command bash -c "
+      set -e
+      # Check if holo-dev-server is in the path
+      if command -v holo-dev-server >/dev/null 2>&1; then
+        echo 'holo-dev-server is available in the PATH while it should not'
+        exit 1
+      else
+        echo 'holo-dev-server is NOT available in the PATH'
+      fi
+    "
+
+    rm -rf /tmp/holo-flake
+    cd /tmp
+
+    hc-scaffold --template vue web-app holo-flake --setup-nix true --holo
+    cd holo-flake
+
+    nix develop --command bash -c "
+      set -e
+      # Check if holo-dev-server is in the path
+      if command -v holo-dev-server >/dev/null 2>&1; then
+        echo 'holo-dev-server is available in the PATH'
+      else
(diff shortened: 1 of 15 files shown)


r-vdp:
@c12i I pushed a commit to hopefully fix the tests. Weird though that github does not show any log output for the failed runs.

run_test.sh
```diff
@@ -148,3 +148,40 @@ set -e
...
+cd /tmp
+
+hc-scaffold web-app holo-flake --template vue
+cd holo-flake
+
+nix develop --command bash -c "
+set -e
+# Check if holo-dev-server is in the path
+if command -v holo-dev-server >/dev/null 2>&1; then
+  echo 'holo-dev-server is available in the PATH while it shouldn\'t'
```
c12i, line 162: The apostrophe doesn't appear to be correctly escaped, might be what's causing CI to fail

```suggestion
  echo 'holo-dev-server is available in the PATH while it shouldn'\''t'
```

c12i (changes requested):

c12i:
> @c12i I pushed a commit to hopefully fix the tests. Weird though that github does not show any log output for the failed runs.

I noticed this too, not sure why this is the case. However, I ran the tests locally (with above fixes), and the script exits with this error

```sh
holo-dev-server is NOT available in the PATH
```

r-vdp:
> > @c12i I pushed a commit to hopefully fix the tests. Weird though that github does not show any log output for the failed runs.
> 
> I noticed this too, not sure why this is the case. However, I ran the tests locally (with above fixes), and the script exits with this error
> 
> ```shell
> holo-dev-server is NOT available in the PATH
> ```

They were definitely working locally for me. I'll look into it.

c12i:
@r-vdp the tests are still failing after your fix, could you kindly have a look? I would like to support, but I'm unfortunately still new to nix, feel free to follow up with the rest of the team on mattermost for assistance with this if it continues to persist

r-vdp:
Hi @c12i. Well, I know why we get the error, but we are still discussing with some other team members about the most desirable manner of fixing it, since there are some trade-offs to take into account.
So I'm waiting for us to take a final decision before continuing on this PR because there's also the possibility that we'll take another approach.

I will mark the PR as draft for now until we've figured out how we want to continue.

r-vdp:
@c12i we finally got all the required pieces in place at the holo side. Would you be able to give this another look?

r-vdp:
Please review the commits separately, as the first two commits are just the result of running standard formatters to avoid reformatting later on.

src/cli.rs
```diff
@@ -674,7 +677,10 @@ Collection "{}" scaffolded!
                     println!("{}", i);
                 }
             }
-            HcScaffoldCommand::Example { example } => {
+            HcScaffoldCommand::Example {
+                example,
+                holo_enabled,
```
c12i, line 682: Thanks for adding this change

c12i (approved):
LGTM, should probably remove it from draft mode, right @r-vdp?

r-vdp:
@c12i I undrafted and pushed a small change to remove a line from the diff that was not needed anymore.

PR #183, opened 2024-01-18
fix(templates): add signal on FooForBar template

c12i:


c12i:
Closing to research and understand why `get_foo_for_bar` zome fns, with an `EntryHash` rather than `ActionHash` results in a deserialization error (`holochain-0.2`)

PR #184, opened 2024-01-18
fix: address file presence in choose_or_get_template options

### LINKED ISSUE #182: Choose UI framework options can contain files present in a `.templates` directory
This bug occurs when scaffolding a web app using a `--templates-path` which contains a local template or when using `--templates-url` which downloads the template to a `.templates` dir.

If the `.templates` folder contains a file. This file will also be included as part of the options when choosing a template, selecting this file causes this error:

```sh
Error: IO error: Not a directory (os error 20)

Caused by:
    Not a directory (os error 20)
```

I believe the expected behaviour would be to skip any files in the `.templates` directory and only have the folder names appear as options

### Minimal reproduction
- run `hc-scaffold templates init` and choose a template
- insert a file in the created `.templates` folder i.e `touch .templates/cat`
- scaffold a web app with the `--templates-path` option set to the path to the `.templates` directory i.e `hc-scaffold web-app --templates-path .templates`
- The `cat` file will appear as an option when choosing a UI framework, select it to reproduce the error above

c12i:
This PR fixes #182

### DIFF
--- a/src/templates.rs
+++ b/src/templates.rs
@@ -250,7 +250,10 @@ pub fn choose_or_get_template_file_tree(
     if dir_exists(file_tree, &templates_path()) {
         let template_name = choose_or_get_template(file_tree, template)?;
 
-        Ok(FileTree::Directory(dir_content(&file_tree, &templates_path().join(template_name))?))
+        Ok(FileTree::Directory(dir_content(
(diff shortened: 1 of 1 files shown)


ThetaSinner (approved):
lgtm!

Issue #173, opened 2024-01-08
Remove `unwrap()` calls and Implement Proper Error Handling


c12i:
There is an abundant presence of `unwrap()` calls in the codebase, which can lead to unexpected panics when errors occur. It is essential to replace these calls with proper error handling mechanisms to ensure robust and reliable software behavior.

We already are using a library `anyhow` that can be useful with this.

PR #185, opened 2024-01-19
refactor: Refactor templates module and handle unhandled errors

c12i:
This PR is an attempt at refactoring some the templates module, I struggled a lot with initially understanding the code and as a result would prove to be challenge in maintaining it, therefore this PR makes the following changes

- Extract the logic inside the for loop into separate functions
- Move regex strings to constants
- Prefer `match` over `if` `else if` for better readablilty
- handle errors gracefully by removing the `unwrap` calls and applying error context via anyhow where necessary, a first step in addressing #173
- perform early returns where possible

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,3 +54,4 @@ colored = "2.1.0"
 dprint-plugin-typescript = "0.91.1"
 dprint-plugin-vue = { git = "https://github.com/c12i/dprint-plugin-vue.git", version = "0.6.0"}
 git2 = "0.19.0"
+once_cell = "1.19.0"
--- a/src/templates.rs
+++ b/src/templates.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use build_fs_tree::serde::Serialize;
 use handlebars::Handlebars;
 use regex::Regex;
@@ -21,6 +22,12 @@ pub mod example;
 pub mod integrity;
 pub mod link_type;
 pub mod web_app;
+
+const EACH_TEMPLATE_REGEX: &str =
+    r"(?P<c>(.)*)/\{\{#each (?P<b>([^\{\}])*)\}\}(?P<a>(.)*)\{\{/each\}\}.hbs\z";
+const EACH_IF_TEMPLATE_REGEX: &str = r"(?P<c>(.)*)/\{\{#each (?P<b>([^\{\}])*)\}\}\{\{#if (?P<d>([^\{\}])*)\}\}(?P<a>(.)*)\{\{/if\}\}\{\{/each\}\}.hbs\z";
+const IF_TEMPLATE_REGEX: &str =
+    r"(?P<c>(.)*)/\{\{#if (?P<b>([^\{\}])*)\}\}(?P<a>(.)*)\{\{/if\}\}.hbs\z";
 
 pub struct ScaffoldedTemplate {
     pub file_tree: FileTree,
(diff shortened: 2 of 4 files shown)


src/templates.rs
```diff
@@ -103,125 +107,178 @@ pub fn render_template_file_tree<'a, T: Serialize>(
...
-                        .into_iter()
-                        .map(|s| s.to_string())
-                        .collect();
-
-                    for (i, f) in files_to_create.into_iter().enumerate() {
-                        let target_path = PathBuf::from(path_prefix.clone()).join(f);
-
-                        transformed_templates
-                            .insert(target_path, Some(new_contents_split[i].clone()));
+            let each_regex = Regex::new(EACH_TEMPLATE_REGEX).context("EACH_TEMPLATE_REGEX is invalid")?;
```
ThetaSinner, line 121: This is better with constants but I'd go a step further and try to avoid recreating the `Regex`. Especially in a loop

src/templates.rs
```diff
@@ -103,125 +107,178 @@ pub fn render_template_file_tree<'a, T: Serialize>(
...
+            let if_regex = Regex::new(IF_TEMPLATE_REGEX).context("IF_TEMPLATE_REGEX is invalid")?;
+
+            match (each_regex.is_match(path_str), if_regex.is_match(path_str)) {
```
ThetaSinner, line 124: I don't think `match` is clearer here, this was a good use of `if`/`else if`/`else`. If nothing else, `match` can't shortcut here where `if`/`else` can

ThetaSinner: Otherwise this refactor is good, but I would keep the `if`/`else`

src/templates.rs
```diff
@@ -103,125 +107,178 @@ pub fn render_template_file_tree<'a, T: Serialize>(
...
+                    if_regex, 
+                    h,
+                    path_str,
+                    &contents,
+                    existing_app_file_tree,
+                    data,
+                    &mut transformed_templates
+                )?,
+                _ => {
+                    if let Some(e) = path.extension() {
```
ThetaSinner, line 145: Somewhere above there was a nice refactor to do this with a `map_or`, same here to avoid the nested `if`?

ThetaSinner (commented):
Really nice! I have a few comments but super happy to see these unwraps going and some more functional style code

src/templates.rs
c12i: Updated

src/templates.rs
c12i: Fixed

Cargo.toml
```diff
@@ -54,3 +54,4 @@ colored = "2.1.0"
 dprint-plugin-typescript = "0.91.1"
 dprint-plugin-vue = { git = "https://github.com/c12i/dprint-plugin-vue.git", version = "0.6.0"}
 git2 = "0.19.0"
+once_cell = "1.19.0"
```
ThetaSinner, line 57: This is now effectively implemented in the standard library. Not something that necessarily needs to be done for this PR, but worth taking a look at https://doc.rust-lang.org/std/cell/

Doesn't cover every use case that this library does though.

ThetaSinner (approved):
Looks good!

Cargo.toml
c12i: Yes I am aware of `cell`, but I think it's use-case would be when dealing with interior mutability and not initializing a shared non-mutable static variable, which is what we are doing here. Kind of similar to `lazy_static`

ThetaSinner: Cool, no problem if it isn't the right tool here. PR is good to go then!

PR #191, opened 2024-01-23
chore: Add support for holochain `0.3`

c12i:
## TODOs

- [x] Bump to latest version of `tryorama` compatible with `0.3` as soon as it's released
- [x] Bump `@holo-host/web-sdk` version (_not really a hard requirement, so this bump will be skipped for now_)
- [x] Smoke test with templates and examples

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,6 +8,7 @@
     holochain = {
       url = "github:holochain/holochain";
       inputs.versions.follows = "versions";
+      inputs.holochain.url = "github:holochain/holochain/holochain-0.3.0-beta-dev.33";
     };
   };
 
--- a/flake.nix
+++ b/flake.nix
@@ -3,12 +3,11 @@
 
   inputs = {
     nixpkgs.follows = "holochain/nixpkgs";
-    versions.url = "github:holochain/holochain?dir=versions/0_2";
+    versions.url = "github:holochain/holochain?dir=versions/weekly";
 
     holochain = {
       url = "github:holochain/holochain";
       inputs.versions.follows = "versions";
-      inputs.holochain.url = "github:holochain/holochain/holochain-0.3.0-beta-dev.33";
     };
   };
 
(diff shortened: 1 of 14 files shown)


flake.nix
```diff
@@ -3,7 +3,7 @@
 
   inputs = {
     nixpkgs.follows = "holochain/nixpkgs";
-    versions.url = "github:holochain/holochain?dir=versions/0_2";
+    versions.url = "github:holochain/holochain?dir=versions/weekly";
```
ThetaSinner, line 6: The `flake.lock` should have changed when this did, have you run `nix flake update`?

src/scaffold/zome.rs
```diff
@@ -340,7 +340,7 @@ pub fn add_common_zome_dependencies_to_workspace_cargo(
         &format!("={}", hdk_version()),
     )?;
     let file_tree =
-        add_workspace_external_dependency(file_tree, &"serde".to_string(), &"=1.0.166".to_string())?;
+        add_workspace_external_dependency(file_tree, &"serde".to_string(), &"1.0.195".to_string())?;
```
ThetaSinner, line 343: This could just be `"1.0"`, it should be fine to let this upgrade now that we're pinning `rmp-serde` which was the reason we ended up pinning here in the first place

ThetaSinner (commented):
Looking good! The version of the scaffolding itself should be updated as part of this PR then I think this is good to go

c12i:
> The version of the scaffolding itself should be updated as part of this PR

The version should be `0.3000.0`, or should I include a beta flag to it?

flake.nix
c12i: Good catch, updated it

ThetaSinner:
It's up to you, it would be reasonable to have `0.3XXX.Y` scaffolding versions based on pre-release versions of Holochain. Or you could keep scaffolding in `-dev` pre-release mode until Holochain 0.3 is released. I think either works :)

ThetaSinner (approved):

PR #196, opened 2024-01-25
enhancement: use `parse(from_os_str)` for `PathBuf` args to handle non UTF-8 file paths on windows

c12i:
This PR introduces a slight modification to the handling of `PathBuf` arguments in the cli. The change involves adding the `parse(from_os_str)` attribute to `PathBuf` fields to enhance the application's compatibility with file paths containing non-UTF-8 characters, particularly on Windows

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -61,7 +61,7 @@ pub enum HcScaffold {
         /// The git repository URL from which to download the template, incompatible with --templates-path
         templates_url: Option<String>,
 
-        #[structopt(short = "p", long)]
+        #[structopt(short = "p", long, parse(from_os_str))]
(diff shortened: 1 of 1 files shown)


ThetaSinner (approved):
Change looks good, any chance of some tests that run on Windows and verify it? It would be the first time we've run tests on Windows for this project so I'd be fine with it being a separate PR but maybe worth doing?

c12i:
> It would be the first time we've run tests on Windows for this project so I'd be fine with it being a separate PR but maybe worth doing?

For now this is going to be done manually, though this is definitely something worth incorporating into our ci/cd workflow with a future PR

PR #199, opened 2024-01-28
fix: pass valid URL to AppAgentWebsocket.connect

mattyg:


c12i (commented):
@mattyg `AppAgentWebsocket.connect` in the `@holochain/client` compatible with holochain `0.1` accepts a string rather than `Url` as its parameter

PR #200, opened 2024-01-29
Use hc-spin cli (develop-0.1)

matthme:
This PR switches to using the hc-spin CLI instead of `hc launch` by default since it provides better developer tools.
It keeps the option to run the happ in tauri explicitly via `npm run start:tauri` if that's desired.

PR #201, opened 2024-01-29
Use hc-spin cli 

matthme:
Switches to using the hc-spin CLI instead of hc launch by default since it provides better developer tools.

### DIFF
--- a/src/versions.rs
+++ b/src/versions.rs
@@ -7,7 +7,7 @@ pub fn holochain_client_version() -> String {
 }
 
 pub fn hc_spin_version() -> String {
-    String::from("^0.200.6")
+    String::from("^0.200.7")
 }
 
 pub fn web_sdk_version() -> String {
--- a/src/versions.rs
+++ b/src/versions.rs
@@ -3,7 +3,11 @@ pub fn tryorama_version() -> String {
 }
 
 pub fn holochain_client_version() -> String {
-    String::from("^0.16.5")
+    String::from("^0.16.7")
+}
+
+pub fn hc_spin_version() -> String {
+    String::from("^0.200.6")
 }
 
 pub fn web_sdk_version() -> String {
--- a/src/versions.rs
+++ b/src/versions.rs
@@ -7,7 +7,7 @@ pub fn holochain_client_version() -> String {
 }
 
 pub fn hc_spin_version() -> String {
-    String::from("^0.200.7")
+    String::from("^0.200.8")
 }
 
 pub fn web_sdk_version() -> String {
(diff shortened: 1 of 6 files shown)


ThetaSinner (approved):

c12i (commented):
Apart from the `--host` fix for `No index.html file found at http://127.0.0.1:8888/index.html` on the `ui/package.json`, everything else works as expected

c12i:
@ThetaSinner I thin this change should only be backported to `develop-0.2`

Since we already have #200 targeting `develop-01`, which I think we should still keep open since the launch scripts are pretty different between `0.1` and `0.2`

ThetaSinner:
That works. Ideally the only difference would be the version of `hc-spin` we're pulling in and yes that would need adjusting on a back-port but that's expected.

If you're happy that the changes are equivalent between the two PRs then that's fine with me. 

For context, the merge to develop then back-port strategy is partly there to ensure we don't lose changes. If bugfixes are submitted directly to the 0.1 and 0.2 branches then it's easy to lose them. That's not a problem here, but in general it's easier to keep track of if general enhancements have a singe place that they are supposed to be made

src/versions.rs
```diff
@@ -3,7 +3,11 @@ pub fn tryorama_version() -> String {
 }
 
 pub fn holochain_client_version() -> String {
-    String::from("^0.16.5")
+    String::from("^0.16.7")
```
c12i, line 6: @matthme is this upgrade needed?

`develop` is currently on `^0.17.0-dev.3`

c12i:
@matthme I haven't tested `hc-spin` with `0.3` yet, and I am not sure whether it's compatible with it too. I think to be on the safe side, we can change the target branch of this PR to `develop-0.2` instead of `develop`. Since `develop` has now been upgraded to be compatible `0.3` (#191) since this PR was last reviewed

cc @ThetaSinner

c12i:
I just noticed that a `0.3` compatible version of client was shipped with the vue.js `cell_id` proxy object fix

I have opened #205 to use the latest version

matthme:
> I haven't tested hc-spin with 0.3 yet, and I am not sure whether it's compatible with it too

Yes, I haven't tested it with 0.3 either and I suspect that I will need to make a dedicated hc-spin version `0.300.x` for that. Changing the target branch to develop-0.2 is fine with me. And I will test hc-spin with 0.3 now that scaffolding is available for it.

c12i:
@matthme, modifying the base branch has introduced commits not present in `develop-0.2` from `develop`, leading to merge conflicts.I recommend reopening this PR against `develop-0.2` to address this issue. What are your thoughts on this approach?

c12i:
Alternatively, I can close this PR and backport #208 into the respective maintenance branches using the correct holochain client version for each

PR #202, opened 2024-02-01
chore: bump holochain dependency versions

c12i:
This PR bumps holochain dependency versions to `0.2.6`


c12i:
@ThetaSinner thanks for the heads up, wonder if that's what might be causing CI to fail

c12i:
Found the root cause of the failing CI, looks like the version requires an update in the Cargo.toml template

ThetaSinner (approved):

PR #203, opened 2024-02-01
feat(cli): add optional flag to skip ui generation for `link-type`, `entry-type` and `collection`

### LINKED ISSUE #186: Option to exclude generating UI code
Really happy to see this repo being active improved! @c12i 

Wanted to share a want I've had for a while: I'd like an option to *not* generate any UI code for the commands `link-type`, `entry-type`, `collection`

c12i:
This adds an optional flag `--no-ui` for the `link-type`, `entry-type` and `collection` commands. This flag skips `ui` generation, though the `tests` folder will still be populated with generated code

This change is non intrusive and does not break the normal behaviour of these commands

closes #186

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -150,6 +150,10 @@ pub enum HcScaffold {
         /// The template to scaffold the dna from
         /// The template must be located at the ".templates/<TEMPLATE NAME>" folder of the repository
         template: Option<String>,
+
+        #[structopt(long, hidden = true)]
+        /// Optional flag to skip ui generation
+        skip_ui: bool,
     },
     /// Scaffold a link type and its appropriate zome functions into an existing zome
     LinkType {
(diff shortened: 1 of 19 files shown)


mattyg (approved):
Just flagging that if a user does *not* generate UI for an entry type, but then *does* generate UI for collection, the collection component will be trying to import an entry type component that is missing.

Not sure what the right solution to this, or if it even matters. This works great for my needs! Thank you!

c12i:
@mattyg thanks for pointing this out, I think it matters, since we want to generate code that runs on the first try

I may have a solution for this case, having a `file_exists` custom handlebars helper to check if a file exists in the file tree before rendering any references to it. This way the references to code that was not generated will not be shown

c12i:
added some accommodation for potentially non existent files that could be referenced @mattyg, feel free to re-review

mattyg (approved):
nice! good approach

PR #205, opened 2024-02-06
chore: update holochain client version

c12i:
This PR updates the holochain client version to the latest `0.3` compatible version that addresses `cell_id` being a proxy object

https://github.com/holochain/holochain-client-js/releases/tag/v0.17.0-dev.5


matthme (approved):
perfect!

PR #207, opened 2024-02-07
Adds holochain_client_version variable to example package.json

matthme:
I haven't actually tested it yet but I'm assuming the example package.json.hbs has access to the `holochain_client_version` variable as well.

### DIFF
--- a/templates/vanilla/example/ui/package.json.hbs
+++ b/templates/vanilla/example/ui/package.json.hbs
@@ -9,7 +9,7 @@
       "package": "npm run build && cd dist && bestzip ../dist.zip *"
     },
     "dependencies": {
-      "@holochain/client": "^0.12.2",
+      "@holochain/client": "{{holochain_client_version}}",
       "@holo-host/identicon": "^0.1.0",
       "@msgpack/msgpack": "^2.7.2"
     },


c12i:
@matthme  unfortunately not, this is the only data passed to the example templates at  https://github.com/holochain/scaffolding/blob/develop/src/templates/example.rs#L24

I do think it makes sense to reference from the `versions.rs` rather than have them hard coded, it becomes very easy to miss updating these hard-coded values

```rs
 let data = ScaffoldExampleData {
        example: example.to_string(),
 };
```

matthme:
Oh I see...yes I agree. It's really easy to miss them (as happened in this case).

templates/vanilla/example/ui/package.json.hbs
```diff
@@ -9,7 +9,7 @@
       "package": "npm run build && cd dist && bestzip ../dist.zip *"
     },
     "dependencies": {
-      "@holochain/client": "^0.12.2",
+      "@holochain/client": "{{holochain_client_version}}",
```
c12i, line 12: Prefer the hardcoded version for now

will create a PR to expose these values to the example templats

c12i (changes requested):

c12i:
Closing in favour of #218

PR #209, opened 2024-02-07
fix(example-lit): fix lit example AppAgentWebsocket connect call

c12i:
A similar change to #199 that was not applied to the example template


ThetaSinner (approved):

c12i:
Just realized #198 was already made against `develop`, the `develop-0.2` branch I ran this change locally with did not have this change backported

I am therefore closing this PR in favour of #198

PR #218, opened 2024-02-19
fix: extend example template data struct

c12i:
This PR extends the `ExampleTemplateData` struct that is provided to the example templates to include dependency versions, unblocks #207


matthme (approved):
Looks good to me! We could also close #207 and add that change to this PR to have everything in one place for backporting. But I'll let that up to you.

c12i:
You're right, moved changes from #207 here too

PR #219, opened 2024-02-19
fix: Update generated README to use newer nix shell command

### LINKED ISSUE #216: Generated README uses old Nix command
This [line](https://github.com/holochain/scaffolding/blob/develop/templates/vanilla/web-app/README.md.hbs#L10) should be `nix develop`. The old command `nix-shell` looks for a `default.nix` which we no longer generate.

c12i:
This PR closes #216


ThetaSinner (approved):
🥳 Thank you!

PR #230, opened 2024-03-04
fix(templates): Add missing imports and types in test templates

c12i:
This PR adds missing imports and types that were producing typescript errors in the `test` workspace of the generated hApp


guillemcordoba (approved):
LGTM

matthme (approved):

PR #243, opened 2024-03-20
fix: prefer dynamic ports for starting ui for local development

c12i:
This change allows for the simultaneous local execution of multiple web happs without encountering port conflicts


matthme (approved):
Looking good! Also just tested it locally.

PR #244, opened 2024-03-20
Update holochain source

guillemcordoba:
I just run `nix flake update`, this unblocks the testing of custom templates with the new nix infrastructure.


c12i (approved):
Thanks, was wondering why I was no

PR #245, opened 2024-03-21
ci: address clippy and fmt errors and add ci checks

c12i:
This PR adds clippy linting and cargo fmt checks on ci and fixes issues detected in the code


.github/workflows/lint-and-format.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@stable
+      
+      - run: cargo fmt --all --check
+
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+     - uses: actions/checkout@v4
```
ThetaSinner, line 24: This checks out the repo and installs rust a second time, should the run step for clippy not be merged into the job above?

src/scaffold/app/nix.rs
```diff
@@ -9,6 +9,7 @@ use crate::error::{ScaffoldError, ScaffoldResult};
 use crate::file_tree::*;
 use crate::versions::holochain_nix_version;
 
+#[allow(clippy::obfuscated_if_else)]
```
ThetaSinner, line 12: Is this a temporary thing pending improving the code? Otherwise a comment explaining the allow would be good :)

src/scaffold/zome/integrity.rs
```diff
@@ -56,134 +55,133 @@ pub fn validate_agent_joining(_agent_pub_key: AgentPubKey, _membrane_proof: &Opt
 ///
 /// You can read more about validation here: https://docs.rs/hdi/latest/hdi/index.html#data-validation
 #[hdk_extern]
-pub fn validate(op: Op) -> ExternResult<ValidateCallbackResult> {{
-    match op.flattened::<(), ()>()? {{
-        FlatOp::StoreEntry(store_entry) => match store_entry {{
-            OpEntry::CreateEntry {{
+pub fn validate(op: Op) -> ExternResult<ValidateCallbackResult> {
```
ThetaSinner, line 58: Nice! Really appreciated to fix some of the lint issues for generated code too. I'm not sure how practical it is to ensure this at every stage but I would appreciate `cargo clippy` passing on newly scaffolded projects most of the time. At least with the Rust toolchain that's provided in Holonix, nothing we can do about clippy getting new lints over time etc :)

src/templates/helpers/filter.rs
```diff
@@ -9,6 +9,7 @@ pub struct FilterHelper;
 /// then uses Handlebars' truthy logic to filter the items in the value.
 /// It also supports the `#if` helper's `includeZero` optional parameter.
 impl HelperDef for FilterHelper {
+    #[allow(clippy::obfuscated_if_else)]
```
ThetaSinner, line 12: Same question here about commenting this lint allow

ThetaSinner (approved):
Epic! It's amazing how fast these little things stack up if you don't have CI checks in place. Great to have them all cleaned up and CI checks added

src/scaffold/app/nix.rs
c12i: I am a little indifferent to this lint rule, clippy recommends using

```rs
let value = if bool_var { "result" } else { "" };
```

instead of:

```rs
let value = bool_var.then_some("result").unwrap_or("");
```

I find the latter much cleaner and _subjectively_ more readable

.github/workflows/lint-and-format.yaml
c12i: I separated them for brevity, though merging them wouldn't also be a bad idea, although logs would need to be checked to know whether it was clippy or fmt that failed

src/scaffold/app/nix.rs
c12i: Fixed, clippy is satisfied with this instead

```rs
let value = bool_var.then_some("result").unwrap_or_default();
```

src/scaffold/zome/integrity.rs
c12i: > I would appreciate cargo clippy passing on newly scaffolded projects most of the time.

Ahh, I had not considered this too, will fix any issues on the generated code

.github/workflows/lint-and-format.yaml
c12i: Fixed, and merged into the `test.yaml` workflow

src/scaffold/app/nix.rs
ThetaSinner: It looks like `rust-clippy#9100` was only added for style reasons which are subjective as you say, but looking at this I'm thinking it's unlikely to compile to the same thing because it's going through an intermediate type to evaluate a boolean condition.

Doesn't seem like a big issue to me, I'd be fine with a comment on the lint rule pointing to this conversation. I do think it's useful to avoid somebody else having to figure out why ignoring it was reasonable by commenting though :)

src/scaffold/zome/integrity.rs
c12i: Fixed issues with generated rust code, clippy should now pass for the zome crates. I would also like to consider programatic formatting of the generated rust code to ensure `cargo fmt` also passes, I found [this code snippet](https://github.com/bcmyers/num-format/blob/v0.1.2/num-format-dev/src/rustfmt.rs) that could be useful with that. Will do this in a separate PR though

PR #252, opened 2024-04-05
feat: add fast track flag to hc-scaffold web-app

### LINKED ISSUE #236: enhancement: Provide option to scaffold `dna` (and `zomes`) on running `hc scaffold web-app`
I would like to propose the option of adding a `--initial-dna` or `--with-dna` or similar to the `hc-scaffold web-app` command to have a scaffolded web app with a dna already pre-scaffolded.

My motivation for this proposal is that `hc-scaffold dna` is itself a very simple command that adds the `zomes` dirs (which are initially empty) and the `dna.yaml` manifest. I believe we can have this initial dna set up early and allow the happ developer to get straight to creating zomes after scaffolding their app, reducing the steps needed to build a hApp.

We can take this a step further by also having the coordinator and integrity zomes already scaffolded for this `dna` if we would like to further cut down the steps needed to get up and running.

Ultimately the `hc-scaffold web-app` command would now have the option to let users also scaffold a `dna` and `zomes`.  They could also skip these options and still run `hc-scaffold dna` and `hc-scaffold zome` like before.

@guillemcordoba @pdaoust, would like to hear thoughts on this.

c12i:
This PR adds a new (and initially optional) `--fast-track` flag to the `hc scaffold web-app` command, introducing a streamlined workflow for Holochain developers. When enabled, this flag automates the scaffolding of the initial dna structure and its associated zome(s). This enhancement is designed to eliminate the need for running separate commands for these initial setup steps, allowing developers to directly proceed to scaffolding entry types and further accelerate their development process.

Closes #236


matthme (approved):
I think it's great. I wonder whether we should add that by default in some fashion. For example when running `hc scaffold web-app` it would ask after finishing the nix stuff `Do you want to scaffold an initial DNA? (y/n)` and then if `y` proceeds as with the `--fast-track` option. And similarly after scaffolding of the dna ask `Do you want to scaffold an initial coordinator/integrity zome pair? (y/n)`.

c12i:
Added the prompt by default, and some extra fixes for some issues I discovered while testing.

PR #253, opened 2024-04-08
revert: Remove file_exists handlebars helper

c12i:
### Summary
This PR proposes the removal of the `file_exists` handlebars helper from the cli and templates. The helper was initially introduced to safeguard against references to non-existent UI component files when the `--no-ui` flag was used during entry-type scaffolding. Its operation required the assumption that the entire hApp file structure existed on the filesystem, a condition met when scaffold commands were executed in a sequential manner (`hc-scaffold web-app` -> `hc-scaffold dna` -> `hc-scaffold zome` -> `hc-scaffold entry-type`).

### Issue
The sequential dependency poses a limitation during example scaffolding processes with `hc-scaffold example`, where the file tree is only written to disk after the entire hApp's components have been defined. In such scenarios, any invocation of `file_exists` within the templates erroneously returns `false`, given the files, while planned in the hApp's structure, do not yet physically exist on the filesystem. This behavior undermines the utility of the `file_exists` helper in dynamic scaffolding contexts and can lead to confusion and errors in hApp development workflows.

### Resolution
The decision to remove, rather than modify or replace the helper or the example scaffolding logic, stems from a desire to simplify the cli and prevent future integration issues as the scaffolding tool evolves. To mitigate the impact of this change, we introduce detailed warnings and recommendations for developers:

- **Warning**: Developers are cautioned about the potential for UI inconsistencies when the `--no-ui` flag is used selectively across different entry types. Specifically, UI elements for entry types not flagged with `--no-ui` might inadvertently expect or reference components from those flagged, leading to integration challenges.
- **Recommendation**: It is advised to apply the `--no-ui` flag consistently across all `hc-scaffold entry-type` commands within a project. This consistent application ensures UI uniformity and circumvents the integration issues highlighted.


src/cli.rs
```diff
@@ -135,7 +135,14 @@ pub enum HcScaffoldCommand {
         fields: Option<Vec<FieldDefinition>>,
 
         #[structopt(long)]
-        /// Skip ui generation, overriding any widgets specified with the --fields option
+        /// Skips UI generation for this entry, overriding any specified widgets in the --fields option.
+        ///
+        /// **WARNING**: Opting out of UI generation for one entry type but not for others that are linked
```
matthme, line 140: I'm assuming this warning is not shown when using the CLI in interactive mode, is it? I.e. when I run `hc scaffold entry-type` and I'm asked "do you want this entry type to be visible in the UI" I might not ever see that warning...

c12i: You're right , the warning is now printed out for the user to see, apart from being visible when using the `--help` flag

matthme (approved):

PR #262, opened 2024-04-15
fix: infer vanilla template for hello world example

### LINKED ISSUE #260: bug: vanillajs should be inferred as the template when scaffolding the hello-world example

c12i:
This PR also adds a check to ensure that the correct template gets used for an example

Closes #260


matthme (approved):
Looking great.

matthme:
Looks like there are some small Rust lint errors though (the test_fmt_lint test failed).

PR #264, opened 2024-04-17
chore: bump holochain client js, tryorama and hc spin versions (0.3)

c12i:


src/versions.rs
```diff
@@ -1,13 +1,13 @@
...
 }
 
 pub fn holochain_client_version() -> String {
-    String::from("^0.17.0-dev.8")
+    String::from("^0.17.0-dev.9")
 }
 
 pub fn hc_spin_version() -> String {
-    String::from("^0.300.2-dev.2")
+    String::from("^0.300.2-dev.3")
```
matthme, line 10: As discussed, let's remove the `^` here as long as holochain 0.3.x is unstable.

matthme (changes requested):

PR #270, opened 2024-04-24
chore: Bump holochain version `0.3.0-beta-dev.48`

c12i:


c12i:
@jost @ThetaSinner, with `tryorama` updated to `v0.16.0-dev.5` I am running into the following issues, hence why the CI workflow continues to fail.

```
[Tryorama - Local Conductor] info: 2024-04-29T08:55:31.797518Z ERROR r2d2: /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-vendor-cargo-deps/c19b7c6f923b580ac259164a89f2577984ad5ab09ee9d583b888f934adbbe8d0/r2d2-0.8.10/src/lib.rs:121: unable to open database file: /private/tmp/nix-shell.eoADxR/T7JpieA-0BLIjKt0NZ6D5/databases/conductor/conductor.sqlite3
```

```
Error: Test timed out in 180000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout"
```

Scaffolded [this repo](https://github.com/c12i/forum-0.3.0-beta-dev.48) here for reproduction. The validation rules `FlatOp::RegisterUpdate` and `FlatOp::RegisterDelete` have been updated like so in the repo since I initially thought the updated validation rules I added in the diffs could have been causing the errors

```rs
        //...
        FlatOp::RegisterUpdate(_update_entry) => Ok(ValidateCallbackResult::Valid),
        FlatOp::RegisterDelete(_delete_entry) => Ok(ValidateCallbackResult::Valid),
        //...
``` 

the errors above still persist despite this.

c12i:
I am also running into this error in the logs while running the `hApp` and tests. Not sure if could also be related to the issue above

```
ERROR holochain::core::ribosome::real_ribosome: crates/holochain/src/core/ribosome/real_ribosome.rs:355: 'cache_lock.write()' exceeded HIGH time threshold elapsed=1.799856458s intervals=[1, 10, 1000]
```

ThetaSinner:
Timing out tests are because app validation is broken at v48 I believe. See https://github.com/holochain/holochain/issues/3660

I can't get my tryorama tests to pass on v48

src/scaffold/entry_type/integrity.rs
```diff
@@ -846,54 +855,62 @@ fn add_entry_type_to_validation_arms(
...
+                                                        let entry = match original_record.entry().as_option() {
+                                                            Some(entry) => entry,
+                                                            None => {
+                                                                if original_action.entry_type().visibility().is_public() {
+                                                                    return Ok(ValidateCallbackResult::Invalid(
+                                                                        "Original record for a delete of a public entry must contain an entry"
+                                                                            .to_string(),
+                                                                    ));
+                                                                } else {
+                                                                    return Ok(ValidateCallbackResult::Valid);
```
matthme, line 880: I'm unsure about this one. I see that the previous scaffolding accepted all deletes other than `OpDelete::Entry` as valid, including for example private entries. But I'm wondering whether that's problematic? I don't know a lot about the workings of a private entry but I imagine that typically it should not be possible to delete a private entry of someone else.

src/scaffold/entry_type/integrity.rs
```diff
@@ -846,54 +855,62 @@ fn add_entry_type_to_validation_arms(
...
-                                                        }
-                                                    }
+                                                        };
+                                                        let original_app_entry = match EntryTypes::deserialize_from_type(
+                                                            app_entry_type.zome_index,
+                                                            app_entry_type.entry_index,
+                                                            entry,
+                                                        )? {
+                                                            Some(app_entry) => app_entry,
+                                                            None => {
```
matthme, line 890: Actually, when looking at the `OpDelete` from before the breaking change, e.g. [here](https://docs.rs/hdi/0.4.0-beta-dev.35/hdi/flat_op/enum.OpDelete.html) I see that there are also other types of entry, namely `Agent`, `CapClaim`, CapClaim` and `PrivateEntry`. Private entry is already covered above but the others aren't. I see that the previous scaffolding all accepted them as valid which also seems problematic to me. But considering them all as invalid like here  be at least as problematic. I just don't understand enough about under which circumstances such other types of entries would make it into validation...
@ThetaSinner do you have any insights here?

src/scaffold/entry_type/integrity.rs
```diff
@@ -846,54 +855,62 @@ fn add_entry_type_to_validation_arms(
...
-                                                                                entry_def.name.to_case(Case::Snake),
-                                                                                entry_def.name.to_case(Case::Snake),
-                                                                                entry_def.name.to_case(Case::Snake)
-                                                                            ).as_str()
-                                                                        )?;
-                                                                        entry_type_match
-                                                                            .arms
-                                                                            .push(new_arm);
-                                                                    }
+                                                if arm.body.as_ref().to_token_stream().to_string().contains("There are no entry types in this integrity zome") {
```
matthme, line 858: Is there maybe another way to know that there are no entry types than relying on a string snippet here (if not, ignore this ocmment)? It just seems a bit prone to accidental failure. It's not obvious that the specific wording of the string is used here so it might easily happen that someone changes that string and things silently fail here.

matthme (changes requested):
Left a few open questions that I don't feel competent to answer myself at the moment.

src/scaffold/entry_type/integrity.rs
```diff
@@ -829,9 +835,12 @@ fn add_entry_type_to_validation_arms(
                                                                         )
                                                                     {
                                                                         let new_arm: syn::Arm = syn::parse_str(
-                                                                             format!(
-"(EntryTypes::{pascal_entry_def_name}({snake_entry_def_name}), EntryTypes::{pascal_entry_def_name}(original_{snake_entry_def_name})) => 
-    validate_update_{snake_entry_def_name}(action, {snake_entry_def_name}, original_action, original_{snake_entry_def_name}),", 
+                                                                             format!(r#"
+EntryTypes::{pascal_entry_def_name}({snake_entry_def_name}) => {{
+    let original_app_entry = must_get_valid_record(action.clone().original_action_address)?;
+    let original_{snake_entry_def_name} = {pascal_entry_def_name}::try_from(original_app_entry)?;
```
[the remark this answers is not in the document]
c12i: > In case it is indeed possible to get different entry types here in validation it might be worth adding a more meaningful error message here as well.

Should a `wasm_error` be emitted or should a `Ok(ValidateCallbackResult::Invalid("message".to_string()))` be returned instead?

src/scaffold/entry_type/integrity.rs
c12i: The `OpDelete` used to be an enum [before](https://docs.rs/hdi/0.4.0-beta-dev.35/hdi/flat_op/enum.OpDelete.html), now it's a [struct](https://docs.rs/hdi/0.4.0-beta-dev.36/hdi/flat_op/struct.OpDelete.html), I assume all the other variants were composed into one a single `OpDelete` struct as per this change https://github.com/holochain/holochain/pull/3627/files#diff-e0f69b27e830196a207906bed69310346863f12e983f2e9ea3acb502e8a12640R149

c12i: @ThetaSinner  I am also very curious about the rationale of removing the generic type `ET` parameter from `OpDelete`, apart from turning the previous enum into a struct

src/scaffold/entry_type/integrity.rs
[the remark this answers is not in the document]
matthme: I'm actually not sure how the implications of the two compare, i.e. I'm not sure how holochain treats a wasm error differently than a Invalid result.

src/scaffold/entry_type/integrity.rs
matthme: Yes right but this seems to make it rather cumbersome to figure out the type of EntryDeletion we're validating. And I'm unclear on how to properly validate types of entries other than ones defined in `EntryTypes`.
@guillemcordoba do you maybe have some thoughts on that? What your rationale was on accepting deletes of all other types of entries than the ones defined in `EntryTypes` as valid?

src/scaffold/entry_type/integrity.rs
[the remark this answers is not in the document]
ThetaSinner: Returning a wasm error will make Holochain retry. Returning Ok with an Invalid message will cause the dependency to be marked as invalid.

src/scaffold/entry_type/integrity.rs
ThetaSinner: The best person to ask about this would be @jost-s, who made the change. 

As for the general discussion about what the validation rules should do here. Ideally not too much really, it's up to the app author what is and isn't valid and they should be writing tests as they go. Any common validation that should always be run should be happening in sys validation anyway really. I haven't had a chance to look too closely at this yet but is this already doing too much?

src/scaffold/entry_type/integrity.rs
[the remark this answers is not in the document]
matthme: Oh okay...so it probably should be marked as invalid? Or is there any failure mode where that deserialization could fail although the entry type is actually correct?

src/scaffold/entry_type/integrity.rs
c12i: > I imagine that typically it should not be possible to delete a private entry of someone else

I suppose this would only be possible if they had a capability claim allowing for this, right? Regardless, I too feel like an invalid result should be returned from this.

I kept it this way since I noticed a similar check is made for `OpRecord::DeleteEntry`

src/scaffold/entry_type/integrity.rs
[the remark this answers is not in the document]
c12i: @matthme also, was this error `Original and updated entry types must be the same` as a result of converting a `Record` into an entry type defined in the `EntryTypes` enum?

c12i:
Refactored the `add_entry_type_to_validation_arms` for better readability

src/scaffold/entry_type/integrity.rs
c12i: You’re correct that without the original entry, it’s unclear what type the deleted entry was. This requires additional code to determine the type. It appears that this is similar to the behavior observed with `OpRecord::DeleteEntry`, similar rationale as well I suppose?

jost-s: Yes, generally for `Op` the original action and entry were removed. This is a performance improvement with the assumption that you would rarely need the original entry when validating updates and deletes. If you do need it in your happ, you can `must_get_entry` or `must_get_action` for the action.

matthme: > The rationale was that it's unlikely you'd want to check the original entry that you're deleting in validation. What do you think?
I don't think it's that uncommon to require the `EntryType`. If you have 4 different `EntryType`s and one of them is not allowed to be deleted, then you need to match on the entry type.

matthme: This is not saying that this optimization is not justified...but it seems difficult to make use of it in the scaffolded boilerplate.

matthme: which would mean that the optimization would not actually make it into scaffolded apps in practice...

jost-s: It's an optimization in Holochain when preparing `Op`s in the app validation workflow.

> This is not saying that this optimization is not justified...but it seems difficult to make use of it in the scaffolded boilerplate.
In the boilerplate you need to do what `flattened` used to do previously, if you want to check different ones of the EntryTypes. `must_get_entry`, then match against the `EntryType` (app, capgrant, ...) and match against the app's entry types.

src/scaffold/entry_type/integrity.rs
matthme: I have to say I'm still not clear on this one. I saw that you changed it to always rejecting deletion if it's a private entry but I'm not sure whether deleting your own private entry might in general an okay (and maybe desirable) thing to do. I also notice that scaffolding does not a way currently to scaffold private entries in the first place I think. So I'm not sure what to do here...

matthme: Well...if scaffolding does not provide a way to scaffold private entries then maybe rejecting deletion of private entries by default is a way of forcing developers that change entry types to be private to add appropriate validation because their app won't work otherwise (and they hopefully notice this during testing).

c12i: > I'm not sure whether deleting your own private entry might in general an okay (and maybe desirable) thing to do. 

This is my reasoning behind my change. 

> I also notice that scaffolding does not a way currently to scaffold private entries

Yes, perhaps having this configured via scaffolding an `entry-type` would guide on how to handle this case appropriately

src/scaffold/entry_type/integrity.rs
```diff
@@ -531,375 +463,311 @@ fn add_entry_type_to_validation_arms(
...
+                            )? {
+                                Some(app_entry) => app_entry,
+                                None => {
+                                    return Ok(ValidateCallbackResult::Invalid(
+                                        "Original app entry must be one of the defined entry types for this zome"
+                                            .to_string(),
+                                    ));
+                                }
+                            };
+                            match original_app_entry {}
```
matthme, line 640: This looks like it's not of any use anymore, is it?

matthme (approved):
There's one more small comment about an unnecessary match statement but I think we can unblock this now. The private entry question is still not fully clear to me but I think it's fair to assume for now that developers using private entries need to write their custom validation logic if they want private entries to be deletable by agents other than the creator of the entry.

src/scaffold/entry_type/integrity.rs
c12i: It's still in use, the code right below it populates the match arm

matthme: Ah I see, this makes sense.

PR #272, opened 2024-04-26
fix: limit examples to non vanilla ui frameworks

c12i:


pdaoust:
curious: what is the driver behind this change? seems like a UX thing that someone has asked for?

c12i:
The `vanilla` option would produce an error when the user selects it when scaffolding a forum example, so it makes sense to just omit it altogether

matthme:
I completely missed this PR, sorry. Does it mean that the hello-world example is not available in vanilla anymore?

c12i:
Hello world is only available in vanilla, forum only limited to the other templates, but not vanilla

matthme:
Alright, that sounds totally reasonable, just wanted to make sure that hello-world is available in vanilla. Making more complex applications in vanilla seems like quite a cumbersome endeavor anyway without much added value I think.

PR #274, opened 2024-05-02
chore: bump holochain version `0.2.8`

c12i:


run_test.sh
```diff
@@ -57,6 +63,18 @@ setup_and_build_happ() {
   hc-scaffold link-type certificate:EntryHash like --delete false --bidirectional false
   hc-scaffold link-type agent:creator post:EntryHash --delete false --bidirectional true
 
+  if [[ -n "$OVERRIDE_HOLOCHAIN_VERSION" ]]; then
+    nix develop --override-input "versions/holochain" github:holochain/holochain/$OVERRIDE_HOLOCHAIN_VERSION --command bash -c "
+      set -e
```
pdaoust, line 68: small non-blocking suggestion: could this sequence of commands and the identical one just below it be DRYed up into a separate function? Same goes for line 98 onwards.

pdaoust (commented):
I don't feel confident enough in this codebase to approve it, and my bash-foo isn't quite good enough to know what the changes in `run_test.sh` are doing, but I did learn some stuff! Presuming the lock files are automatically updated and can be trusted to be correct -- the commit hash is at least correct :)

matthme (approved):
Nice! Is the override argument useful to test releases that are not yet in holonix?

c12i:
Yep, and to also allow automatic testing of new releases as they get rolled out. I'm setting up a [cron job](https://github.com/c12i/scaffolding/blob/add-release-check-cron/.github/workflows/check-new-releases.yaml) similar to what you have in your [holochain-binaries](https://github.com/matthme/holochain-binaries) that will be running regular checks

Issue #167, opened 2023-12-20
Running `hc scaffold collections` removes code comments


c12i:
Running `hc scaffold collections` deletes code comments from previously generated rust code in integrity zome

### How to reproduce

1. Scaffold web app, name it, select a template of your choice and set up holonix development environment. Cd into the project and run `nix develop`

```sh
nix run github:/holochain/holochain#hc-scaffold -- web-app
```

2. Scaffold a dna

```sh
hc scaffold dna
```

3. Scaffold a zome

```sh
hc scaffold zome
```

4. Scaffold an entry type

```sh
hc scaffold entry-type
```

5. Commit changes via git to visualize deleted comments via git diff

```sh
git add .
git commit -m "Initial commit"
```

7. Scaffold a collection

```sh
hc scaffold collection
```

8. Observe diff, the integrity zome's code will show comments initially scaffolded in previous stages are now deleted. Specifically on the `validate_create_link_<name>_updates` function

PR #281, opened 2024-05-13
fix: prevent unaffected rust files from getting formatted

### LINKED ISSUE #261: Disable reformatting of unaffected files
It seems that the scaffolding tool re-formats files even if they have been unaffected by a scaffolding step. For example if I create two entry types, then I add my personal inline-comments in a file about entry type A and later scaffold a collection for entry type B, my personal inline-comments for entry type A get removed by the scaffolding tool.
Also if my IDE changed the formatting, files get re-formatted by the scaffolding tool in conflicting ways. In that way it is really cumbersome to distinguish which changes are merely formatting changes (or removal of my own inline comments) and where there are actual code changes.
This is especially inconvenient if one adds custom code logic in between scaffolding steps and wants to make sure nothing of one's own code logic has been affected by a scaffolding step.
It would be great to at least not have the scaffolding tool re-format files that are unaffected by a scaffolding step in regards of actual code logic.

c12i:
This PR fixes #261

With regards to the scaffolding tool removing comments, I think it's easier if this is tackled in a separate PR. I have re-opened the original issue to keep track of it as well #167 since the root cause is now known.

### DIFF
--- a/src/file_tree.rs
+++ b/src/file_tree.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use build_fs_tree::{dir, file, Build, FileSystemTree, MergeableFileSystemTree};
 use ignore::WalkBuilder;
 use include_dir::Dir;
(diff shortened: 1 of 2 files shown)


matthme (approved):
Nice!

PR #304, opened 2024-06-05
chore: bump weekly

c12i:


ThetaSinner (approved):
Bump looks good, build is unhappy. Any further with diagnosing that problem?

c12i:
Yea I have looked into this. It's actually the issue I reported on MM with `tryorama`'s `dhtSync` function. Narrowed down the problematic code to [here](https://github.com/c12i/tryorama/blob/bb6eadb367034b5e2beb3a719b00cf24a21ca819/ts/src/util.ts#L76). Not sure if a new version of holochain caused this regression.

Replacing `dhtSync` with `pause` fixes the failing tests though. Do you think I should switch to this for now?

matthme (approved):

matthme (approved):

c12i:
Merging this for now, @matthme has already opened the `dhtSync` issue with `tryorama` 

https://github.com/holochain/tryorama/issues/211

PR #312, opened 2024-06-13
chore: bump versions

c12i:


src/versions.rs
```diff
@@ -1,8 +1,8 @@
 /// npm: https://www.npmjs.com/package/@holochain/tryorama
-pub const TRYORAMA_VERSION: &str = "^0.17.0-dev.0";
+pub const TRYORAMA_VERSION: &str = "^0.17.0-dev.1";
 
 /// npm: https://www.npmjs.com/package/@holochain/client
-pub const HOLOCHAIN_CLIENT_VERSION: &str = "^0.18.0-dev.2";
+pub const HOLOCHAIN_CLIENT_VERSION: &str = "^0.18.0-dev.3";
 
 /// npm: https://www.npmjs.com/package/@holochain/hc-spin
 pub const HC_SPIN_VERSION: &str = "^0.400.0-dev.1";
```
matthme, line 8: You might also use `0.400.0-dev.2` here that I just published today (it shouldn't matter though since the latest will get picked anyway). I will approve the PR anyway, it looks good otherwise.

matthme (approved):

PR #314, opened 2024-06-19
chore: backport `0.3` and bump versions

c12i:


src/versions.rs
```diff
@@ -1,11 +1,11 @@
...
 /// npm: https://www.npmjs.com/package/@holochain/client
-pub const HOLOCHAIN_CLIENT_VERSION: &str = "^0.17.0-rc.0";
+pub const HOLOCHAIN_CLIENT_VERSION: &str = "^0.17.0";
 
 /// npm: https://www.npmjs.com/package/@holochain/hc-spin
-pub const HC_SPIN_VERSION: &str = "0.300.2-dev.6";
+pub const HC_SPIN_VERSION: &str = "0.300.3";
 
 // TODO: update to 0.3 compatible version
 /// npm: https://www.npmjs.com/package/@holo-host/web-sdk
```
ThetaSinner, line 11: I can't comment on it here, but the HOLOCHAIN_HOLONIX_VERSION below should be changed from `0_3_rc` to `0_3`

c12i: Good catch, updated

ThetaSinner (approved):
Looks good!

PR #313, opened 2024-06-17
refactor: clean up duplicated templates

c12i:
This PR cleans up the `templates` dir by grouping duplicated templates (specifically the `tests` folders) across the in built ui frameworks into a separate generic template and handle merging of the contents of these templates when getting the template filetree.

Additionally this PR also moves code related to scaffold config into a separate module in `src/scaffold/config.rs`

Closes #308

PR #316, opened 2024-06-20
feat: make scaffolding cli package manager agnostic

### LINKED ISSUE #268: feat: Make scaffolding cli package manager agnostic
Currently, scaffolding ~only~ implicitly supports `npm` as the package manager when working the hApp's workspace. If the developer has a preference for a different package manager, they would need to add support for it manually, updating scripts that run scripts and add dependencies to workspace members.

This issue suggests making the cli package manager agnostic, meaning, providing the user the option to choose which package manager they would want to use while scaffolding their hApp i.e `yarn`, `pnpm` or `bun`, then ensuring all references to this script manager are used elsewhere.

c12i:
Closes #268 

~Depends on #313~


README.md
```diff
@@ -1,6 +1,37 @@
 # Holochain Scaffolding CLI
 
-CLI to easily generate and edit holochain apps.
+A command-line interface for creating and modifying a Holochain application (hApp).
```
c12i, line 3: @pdaoust I'd like to know whether this doc should be in the developer reference rather than here in the scaffolding cli?

matthme (changes requested):
Hmm...it seems to me that the `--package-manager` option should not be at the top level but instead be an option of the `web-app` and `example` subcommands as it does not seem to have any effect beyond the initial web-app scaffolding step.
And in case that makes sense to you then I think it would be quite nice to have an interactive selection step where one can choose the package manager explicitly, akin to how the UI framework can be chosen.

c12i:
Good idea, thought I would give it the same treatment as the `--template` flag, but you are right, the option is only relevant to `hc-scaffold web-app` and `hc-scaffold example`. Fixed.

README.md
pdaoust: 😬 that is a very good question for which I have no answer. Ideally we'd have documentation 'wherever the dev finds themselves', so I would like to say 'yes to both places'. But that creates duplicated content. I wonder if we could write something that would inline it into the `docs-pages` build pipeline.

templates/custom-template/custom-template/template/web-app/package.json.hbs
pdaoust, line 1: Does this file and the one below it in the `ui` folder also need the package manager treatment?

templates/headless/web-app.instructions.hbs
```diff
@@ -3,16 +3,16 @@ This is a headless skeleton for a hApp.
 To build the app you can run
 
   cd {{app_name}}
-  npm install
-  npm run build:happ
+  {{(package_manager_command package_manager "install" null)}}
```
pdaoust, line 6: ~~it doesn't look like the headless one gets a `package.json` for this command to run against. Should it? it'd be convenient to have the `build:happ` command etc.~~

Wait, I'm guessing that it uses the `package.json.hbs` template from the `generic` template?

pdaoust (commented):
I love all the new documentation! Just one question that may be important. Don't know whether that should be a 'comment' or 'request changes'.

templates/headless/web-app.instructions.hbs
c12i: Yes, it inherits from the generic template

templates/custom-template/custom-template/template/web-app/package.json.hbs
c12i: No, this is a template for creating a custom template, where the author has the freedom to pick which package manager they would wan to use

README.md
c12i: We can publish this through Rust docs and then add a link to it in the developer guide, what do you think?

matthme (approved):
Great job! This PR is a nice improvement :)

PR #317, opened 2024-06-28
refactor: cli module

c12i:
This PR refactors the `cli.rs` module by extracting each subcommand into its own submodule for improved maintainability and modularity


src/cli/web_app.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        build_file_tree(file_tree, &app_folder)?;
+
+        if setup_nix {
+            if let Err(err) = setup_nix_developer_environment(&app_folder) {
+                fs::remove_dir_all(&app_folder).await?;
+                return Err(err)?;
+            }
+        }
+
```
matthme, line 95: It looks like the fast-track part got remove here, is this on purpose? I liked it, if it's possible to add it again I think it would be a plus.

src/cli/entry_type.rs
```diff
@@ -0,0 +1,114 @@
...
+            }
+            None => input_with_case("Entry type name (snake_case):", Case::Snake)?,
+        };
+
+        let dna_file_tree = DnaFileTree::get_or_choose(file_tree, self.dna.as_deref())?;
+        let zome_file_tree =
+            ZomeFileTree::get_or_choose_integrity(dna_file_tree, self.zome.as_deref())?;
+
+        if self.no_ui {
+            let warning_text = r#"
```
matthme, line 75: I did not see this text being printed out at all when I tested it with the `--no-ui` flag. Also, it looks like the interactive choice of whether or not to scaffold the entry type with UI comes only later, as part of the `scaffold_entry_type()` function if I'm not mistaken. So maybe the warning should get moved into that function instead. And I think it would be somewhat important that there is still the possibility to cancel the scaffolding of the entry-type at the moment this warning appears.

matthme (changes requested):

src/cli/web_app.rs
c12i: Good catch, this was accidental, the change was stashed and not popped

src/cli/entry_type.rs
c12i: mmmh, wasn't able to reproduce it, the text get's printed on my end.

But I see the value of having the warning moved to the `scaffold_entry_type` function

matthme: Oh I see :+1:

matthme (approved):

src/cli/entry_type.rs
matthme: Actually, in light of this misunderstanding I'm starting to think that the warning should be moved outside of the function again and be displayed even before asking the entry name on [this line](https://github.com/c12i/scaffolding/blob/9ad5fd5e025708082cd8b4b54c435eda9dfc0ce9/src/cli/entry_type.rs#L62).
And on my Ubuntu terminal the warning is more brownish than yellow and therefore seems more like a comment, maybe `bright_wellow()` would provide a bit more contrast? It's possible that the warning was actually printed and I just didn't notice it because it did not appear at the moment when I expected it (I expected it to appear in the interactive part where it asks whether the entry field should be shown in the UI).
image

PR #320, opened 2024-07-02
chore: dependency upgrades

c12i:


src/scaffold/app/utils.rs
```diff
@@ -43,7 +43,7 @@ pub fn get_or_choose_app_manifest_path_for_dna_manifest(
     Ok(path)
 }
 
-fn bundled_dnas_paths(
+pub fn bundled_dnas_paths(
```
ThetaSinner, line 46: Related change?

ThetaSinner (approved):
Looks good! Glad there's a replacement for `serde_yaml`

matthme (approved):

src/scaffold/app/utils.rs
c12i: Not really , good catch

PR #321, opened 2024-07-08
feat: standardize template styling

### LINKED ISSUE #351: Standardize template styling
Migrate `vue`, `svelte`, and `lit` templates to new styling scheme. Drop material ui.

c12i:
This PR standardizes the styling for scaffolding templates, moving away from Material UI and using CSS to match the recently added React template.

Closes #351 

### CI Failures

~CI failures are due to a recent issue with `@holochain/client`, which will be addressed in PR #341.~

### Changes

- **Styling**: Transitioned from Material UI to CSS for consistency with the new React template.
- **Vec<T> UI Generation**: Removed UI generation for `Vec<T>` field types. The `vec` cardinality on existing field types was not initially considered, and managing this adds complexity. Developers are now expected to implement their own UI for `Vec<T>` fields.


templates/ui-frameworks/lit/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{kebab_case collection_name}}.ts.hbs
```diff
@@ -51,29 +51,29 @@ export class {{pascal_case collection_name}} extends LitElement {
...
-        </div>
-      `,
+      pending: () => html`<progress></progress>`,
       complete: (links) => this.renderList([...this.signaledHashes, ...links.map(l => l.target)]),
-      error: (e: any) => html`<span>Error fetching the {{lower_case (plural referenceable.name)}}: ${e.message}.</span>`
+      error: (e: any) => html`<article>Error fetching the {{lower_case (plural referenceable.name)}}: ${e.message}.</article>`
     });
   }
+
+  static styles = css`${unsafeCSS(sharedStyles)}`;
```
matthme, line 78: I think in lit the more common way to do this would be to define the styles in a `.ts` file that exports the styles as a `css` template string instead of a `.css` file but I guess this would mess up with the reusability of the `index.css` file?

templates/ui-frameworks/lit/link-type/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{#if (and bidireccional (and to_referenceable (ne from_referenceable.hash_type 'AgentPubKey')))}}{{kebab_case (plural from_referenceable.name)}}-for-{{kebab_case to_referenceable.name}}.ts{{¡if}}.hbs
```diff
@@ -46,11 +47,11 @@ export class {{pascal_case (plural from_referenceable.name)}}For{{pascal_case to
   }
 
   renderList(hashes: Array<{{to_referenceable.hash_type}}>) {
-    if (hashes.length === 0) return html`<span>No {{lower_case (plural from_referenceable.name)}} found for this {{lower_case to_referenceable.name}}</span>`;
-    
+    if (!hashes.length) return html`<article>No {{lower_case (plural from_referenceable.name)}} found for this {{lower_case to_referenceable.name}}</article>`;
```
matthme, line 50: Is this an `<article>` tag on purpose?

matthme: According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) for an article tag it's used for content that's "intended to be independently distributable or reusable" like for example a full forum post. I'm not sure this would apply here...

templates/ui-frameworks/lit/web-app/ui/package.json.hbs
```diff
@@ -19,18 +19,7 @@
   {{/if}}
     "@lit-labs/context": "^0.2.0",
     "@lit-labs/task": "^2.0.0",
-    "@material/mwc-circular-progress": "^0.27.0",
```
matthme, line 22: :partying_face: I like seeing these get removed ;)

templates/ui-frameworks/lit/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{kebab_case collection_name}}.ts.hbs
c12i: Yep it would hinder reusability. I could still create one but I would still need to import it in every file and import the css file into it but I'd still end up using the `unsafeCSS` function to do that.

templates/ui-frameworks/lit/link-type/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{#if (and bidireccional (and to_referenceable (ne from_referenceable.hash_type 'AgentPubKey')))}}{{kebab_case (plural from_referenceable.name)}}-for-{{kebab_case to_referenceable.name}}.ts{{¡if}}.hbs
c12i: Yes, this was a misused semantic tag, replaced it with a `div`

templates/ui-frameworks/lit/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{kebab_case collection_name}}.ts.hbs
matthme: With a `css` template string you'd not need to use `unsafeCSS`. But I see that it hinders code reusability across UI frameworks.
Another possible approach that would I think be a bit prettier from an optics standpoint would be to have a file `sharedStyles.ts` in the lit project (next to index.css) with the following content:
```
import { css, unsafeCSS } from "lit";
import styles from "./index.css";

export const sharedStyles = css`${unsafeCSS(styles)}`;
```
and then in the components where you need the styles you'd just import the styles from that file:
```
import { sharedStyles } from "./sharedStyles";

[...]

  static styles = sharedStyles;

````

Comment (matthme) [reply templates/ui-frameworks/lit/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{kebab_case collection_name}}.ts.hbs:78]: So you would only use the `unsafeCSS` function one single time when re-exporting them in the `sharedStyles.ts` file and it would not show up in all the components repeatedly.

File: templates/ui-frameworks/svelte/web-app/ui/index.html.hbs [hunk 4]
```diff
@@ -5,13 +5,11 @@
     <link rel="icon" type="image/svg+xml" href="/vite.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Svelte hApp</title>
-
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons"
```
matthme, line 8: There's still a material link left over here. And another one in `templates/custom-template/custom-template/template/webapp/ui/index.html.hbs`

templates/ui-frameworks/lit/collection/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{kebab_case collection_name}}.ts.hbs
c12i: Good idea, I have also been contemplating another approach where I to create a class that extends `LitElement` i.e `BaseElement`, inside this class I apply the shared styles then for all other components in the app I extend the `BaseElement` rather than the `LitElement`

```ts
export class BaseElement extends LitElement {
  static styles = [
    unsafeCSS(css),
  ];
}

@customElement('holochain-app')
export class HolochainApp extends BaseElement {
  render() {
    return html`...`;
  }
}
```

What do you think?

c12i: I think the method you suggest is much cleaner though, and doesn't involve changing too much code

matthme: Your solution is interesting as well - I haven't thought of that. I don't know what's better...

matthme (approved):

PR #329, opened 2024-07-11
Use latest Nix

ThetaSinner:
Doesn't fix anything, just following what is happening in other repos to move to a recent Nix

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,13 +37,13 @@ jobs:
         uses: ./.github/actions/extend-space
   
       - name: Install nix
-        uses: cachix/install-nix-action@v25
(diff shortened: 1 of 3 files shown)


c12i (changes requested):
Perhaps the template's workflow could also get this update too while you are at it

https://github.com/holochain/scaffolding/blob/28a5afa39e80ba3a264ed510cac0776f695ef72d/templates/generic/web-app/.github/workflows/test.yaml.hbs

ThetaSinner:
Ahh yes, good point!

Flakes are also now enabled by default in Nix so I've removed the extra lines to configure those. Those should go from the Holochain setup too actually

c12i (approved):

PR #343, opened 2024-07-30
chore: backport and bump versions 0.3.2

c12i:


ThetaSinner (approved):

c12i:
@ThetaSinner I am still encountering [this error](https://github.com/holochain/scaffolding/actions/runs/10168869904/job/28124489779?pr=343#step:6:3869) while using `0_3` holonix version. Should I keep using `0_3_rc` for this release?

ThetaSinner:
Hmm, this is the trouble with the circular dependency. You need a newer Holochain for the tests to pass but once it's merged you want to use `0_3`. We should ideally look at tweaking the workflow to override the version of Holochain to use. That way you can test with a different combination

ThetaSinner:
I'm not really sure we want to invest that effort now though, when we're about to mess with Holonix. So maybe the best solution for today would be to release scaffolding twice? Bump it with 0_3_rc, then we can release Holonix and then you can release scaffolding again on 0_3. Not a very tidy solution, but it'll work I think?

c12i:
Yea, that sounds fair, no problem releasing another version later on

PR #350, opened 2024-08-05
Update holonix version to 0_3 instead of 0_3_rc

guillemcordoba:


c12i (approved):
Approved. @ThetaSinner, I assume the `Error while importing \"env\".\"__hc__close_chain_1\": unknown import` has been addressed?

PR #354, opened 2024-08-12
feat: format with rustfmt if available

### LINKED ISSUE #342: fix: Where possible exec `cargo.fmt` on rust code generation
The current code generation process uses [prettyplease](https://github.com/dtolnay/prettyplease) to ensure the Rust code generated from scaffolding is well-formatted. However, leveraging rustfmt for formatting would be more ideal. While rustfmt's API is not publicly available and requires some workarounds to integrate directly, it's reasonable to assume that most Rust toolchains include the rustfmt component. Consequently, cargo fmt can be invoked directly from the scaffolding using the [Command](https://doc.rust-lang.org/std/process/struct.Command.html) struct.

c12i:
Closes #342

### DIFF
--- a/src/cli/collection.rs
+++ b/src/cli/collection.rs
@@ -13,7 +13,7 @@ use crate::{
         zome::ZomeFileTree,
     },
     templates::ScaffoldedTemplate,
-    utils::{check_case, input_with_case},
+    utils::{check_case, input_with_case, run_cargo_fmt_if_available},
 };
 
 #[derive(Debug, StructOpt)]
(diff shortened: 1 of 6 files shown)


matthme (approved):
That's fabulous, I just tested it! No reformatting when I save with my IDE. The only thing I'm wondering whether it might add additional value or not would be to add a `rustfmt.toml` file to be explicit about some of the formatting rules. I'm not sure about that but I've been wondering whether people with special IDE setups might still run into reformatting issues if such a file is not present.

c12i:
No issues running without this file—`rustfmt` defaults will apply. I'd prefer keeping it this way, allowing devs to add their own `rustfmt.toml` if they want custom rules.

PR #355, opened 2024-08-12
feat: run test-cases concurrently

c12i:


c12i:
Closing since no significant runtime performance increase is achieved with this potential change

PR #362, opened 2024-09-03
chore: Switch to `github:holochain/holonix`

c12i:
Updates Nix flake for repo and scaffolded happs to use `github:holochain/holonix`. 

Key changes:
- Scaffolding CLI now exposed as a Nix package, installed in `default` and `ci` devShells
- Eliminates workarounds from the [manual testing guide](https://github.com/holochain/scaffolding/blob/develop/guides/manual-testing.md)
- Retains most functionality from previous flake

Note: Merge #321 first to resolve Svelte template CI failure.


.github/workflows/test.yaml
```diff
@@ -14,17 +14,35 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Extend space
+        uses: ./.github/actions/extend-space
```
ThetaSinner, line 18: Is there a reason for checking this in? I've been able to use a public one with reasonable success elsewhere

.github/workflows/test.yaml
```diff
@@ -14,17 +14,35 @@ jobs:
...
+
+      - name: Install nix
+        uses: cachix/install-nix-action@v27
+        with:
+          install_url: https://releases.nixos.org/nix/nix-2.23.2/install
```
ThetaSinner, line 23: Can move to 24.4 if you want, Holonix has moved. Up to you though!

.github/workflows/test.yaml
```diff
@@ -14,17 +14,35 @@ jobs:
...
+          extra_nix_config: |
+            experimental-features = flakes nix-command
```
ThetaSinner, line 25: No longer required? these should be enabled by default

flake.nix
```diff
@@ -1,77 +1,144 @@
...
 
-    holochain = {
-      url = "github:holochain/holochain";
-      inputs.versions.follows = "versions";
+    nixpkgs.follows = "holonix/nixpkgs";
+    flake-parts.follows = "holonix/flake-parts";
+
+    crane = {
+      url = "github:ipetkov/crane";
+      inputs.nixpkgs.follows = "holonix/nixpkgs";
```
ThetaSinner, line 12: Should not be required if you do a flake update, they have removed their nixpkgs input

flake.nix
```diff
@@ -1,77 +1,144 @@
...
+    };
+
+    rust-overlay = {
```
ThetaSinner, line 15: Holonix will also be pulling in the rust-overlay and crane, one should follow the other.

I am currently preferring that dependencies which use the same inputs as you should be made to follow your inputs - see Wind Tunnel or tryorama for the way the follows have been set up there.

This is optional but I think it's cleaner to use one toolset

flake.nix
```diff
@@ -1,77 +1,144 @@
...
-      }: {
-        devShells.default = pkgs.mkShell {
-          inputsFrom = [inputs'.holochain.devShells.rustDev];
-          packages = [pkgs.nodejs_20];
-        };
+  outputs = inputs @ { flake-parts, nixpkgs, crane, rust-overlay, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; }
+      rec {
+        flake = {
+          templates.default = {
```
ThetaSinner, line 25: Nice!

flake.nix
```diff
@@ -1,77 +1,144 @@
...
+                ++ (lib.optionals pkgs.stdenv.isDarwin
+                (with pkgs.darwin.apple_sdk.frameworks; [
+                  CoreFoundation
+                  SystemConfiguration
+                  Security
+                ]));
 
+              nativeBuildInputs = [ pkgs.perl ];
+              env = {
+                OPENSSL_DIR = "${pkgs.openssl.dev}";
```
ThetaSinner, line 100: Does this fail to build without openssl? I'd like to take a look at why with you if possible. Probably not right now because I'm busy with some Holochain work but I'd managed to eliminate the need for this dependency and I'd like to figure out why it's needed again! :)

templates/custom-template/custom-template/flake.nix
ThetaSinner, line 1: Should this template use the new Holonix too?

ThetaSinner (commented):
Looks great! I've left quite a few comments but they're all pretty minor things and several are optional :)

flake.nix
c12i: Seems to work now without openssl, but I need to include `perl`

```nix
 nativeBuildInputs = [ pkgs.perl ];
```

.github/workflows/test.yaml
c12i: Yes, no need for this, replaced with a public one

templates/custom-template/custom-template/flake.nix
c12i: Updated

flake.nix
c12i: I see, I'll leave it as is for now, will update once I have a better understanding of how the new holonix is set up.

ThetaSinner: Cool, yeah it's nothing urgent. Happy to talk it through on a call because it took me a while to get my head round why I was seeing people do this in public flakes

PR #371, opened 2024-09-18
docs:  update manual testing docs

c12i:


guides/manual-testing.md
```diff
@@ -1,37 +1,59 @@
...
-```shell
-nix develop --override-input "versions/scaffolding" <path-to-local-scaffolding-clone>
-```
+4. Scaffold a new app:
+
+   ```shell
+   hc-scaffold web-app
+   ```
+
+5. Follow the Scaffolding tool's output instructions, but skip the `nix develop` step to ensure you continue using the locally built `hc-scaffold`.
```
ThetaSinner, line 28: Doesn't this mean you'll be missing package managers etc?

guides/manual-testing.md
```diff
@@ -1,37 +1,59 @@
...
 
-Which will ensure that you keep using the updated Scaffolding tool inside the scaffolded hApp environment.
+6. Run tests or launch your hApp using `nix develop` as needed.
```
ThetaSinner, line 30: Oh I see, so you'd do your scaffolding in the shell that has the custom hc-scaffold but use another shell to do your app commands - could that be made more explicit in the steps?

ThetaSinner (approved):
Nice! Looks good. I think just one place where it can be a bit more explicit :)

PR #372, opened 2024-09-18
chore: update 0.4 flake inputs

c12i:


ThetaSinner (approved):
Looks good! We'll have to communicate when we want to move nixpkgs versions to keep caching working but I think this is a good pattern :)

PR #378, opened 2024-09-23
revert to yarn v1.22.22 nixpkg

c12i:
For existing hApps, the node_modules dir(s) will need to be deleted and this command ran while in the nix shell:

```sh
corepack prepare yarn@1.22.22 --activate
```


matthme (approved):
Looks good. I think it makes sense to revert for now. The solution I thought might work for hc-spin turned out to not work so I will have to spend more time trying to come up with a solution.

PR #380, opened 2024-09-24
feat: Create links from/to `ExternalHash`

### LINKED ISSUE #254: [Feature] Link-types from AnyLinkableHash to AnyLinkableHash
Currently `hc scaffold link-type` only allows to create links from entry types to entry types, each of which must already have been created. A link in general however can point from an `AnyLinkableHash` to an `AnyLinkableHash`. A common use case for me is for example to make a link to or from an `AgentPublicKey`, I would like to have that be supported.

c12i:
Closes #254
Closes https://github.com/holochain/scaffolding/issues/389


src/error.rs
```diff
@@ -54,8 +54,8 @@ pub enum ScaffoldError {
     #[error("Folder already exists: {0}")]
     FolderAlreadyExists(PathBuf),
 
-    #[error("Invalid reserved word: {0}")]
-    InvalidReservedWord(String),
+    #[error("Invalid use of reserved {context} keyword '{word}'.in this context.")]
```
matthme, line 57: > '{word}'.in this

Is there an accidental dot here?

src/scaffold/entry_type/utils.rs
```diff
@@ -50,6 +41,7 @@ fn inner_choose_referenceable(
     all_options.push("Agent".to_string());
 
     if optional {
+        all_options.push("AnyLinkableHash".to_string());
```
matthme, line 44: I would expect the `AnyLinkableHash` to also be available in the non-optional scenario, or is there a reason for it not to be available?

matthme:
I noticed that the `[None]` option provided by `get_or_choose_optional_reference` does not make sense to me (I see that this was already there before this PR). [Here](https://github.com/holochain/scaffolding/blob/d36abccbfbff833a8eb73b38be848856b0d38f53/src/scaffold/link_type.rs#L50-L54), why would a link only optionally point to a target? The actual behavior now when a link-type is scaffolded with target `[None]` is that the target is the same as the base which seems not useful in any real case I can think of. It seems to me that for link-types, both base and target should always be non-optional.

And as another remark, I think it might be worth adding `ExternalHash` as an explicit option as well. `AnyLinkable` hash is the more generic type which entails `ActionHash`, `EntryHash` and `ExternalHash` but sometimes it might be desirable to be explicit that one wants to point to an `ExternalHash`.

matthme:
It might even make sense to replace `AnyLinkableHash` with `ExternalHash` since if all `ActionHash`, `EntryHash`, `AgentPubKey` and `ExternalHash` are available as options then `AnyLinkableHash` is covered implicitly as well.

matthme:
> It might even make sense to replace AnyLinkableHash with ExternalHash

I'm not sure though...maybe there are cases where this flexibility is desired, i.e. that you'd want to allow for a link that can point to either of these types and is not hard-coded. So keeping `AnyLinkableHash` as an option as well (alongside `ExternalHash` and the others) might not hurt.

src/scaffold/entry_type/utils.rs
c12i: The reason that it's not available is to only allow setting an `ExternalHash` (prev `AnyLinkableHash`) as a target rather than base. This just for semantical reasons really, what do you think?

c12i:
> I noticed that the [None] option provided by get_or_choose_optional_reference does not make sense to me (I see that this was already there before this PR)

Yes this `[None]` option has always existed, it's exact usecase is actually not documented, perhaps @guillemcordoba can shed some light. Decided to keep it to not potentially break anything that might depend on it.

> why would a link only optionally point to a target?

From my understanding of the previous code, I think the term optional doesn't really communicate the purpose correctly. I see that the `optional` check just adds more options to the default list of all entries that can have links created to it. The difference between `get_or_choose_referenceable` and `get_or_choose_optional_reference_type` fundamentally ends up being that the latter contains extra options i.e `None` and `ExternalHash` that are not provided by `get_or_choose_referenceable` to make sure that these two options are only assignable to the "to referenceable"

guillemcordoba:
The `[None]` option is to be able to support "metadata" only links. In holochain, the only way to attach a small piece of extra information to a hash is creating a link from that hash and add the information as the tag. In this case the target is actually irrelevant, because the only point of the link is to attach is link tag to the hash. This is the use case I wanted to support with the None option.

You might argue that this use case could be done by targeting any other entry hash. But then the scaffolding tool would scaffold validation to make sure that the target entry exists, which is not what I want if I only want to attach metadata to that hash.

Hope this helps!

src/scaffold/entry_type/utils.rs
matthme: I think there are also cases where you might want to have an external hash as a base...

matthme:
Oh interesting, yes this helps. Maybe that could be described a bit more clearly in the option then. Something like `[None] (use this link to attach meta-data only)` or something.

templates/generic/link-type/tests/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{#if to_referenceable}}{{kebab_case from_referenceable.name}}-to-{{kebab_case (plural to_referenceable.name)}}.test.ts{{¡if}}.hbs
```diff
@@ -43,21 +48,34 @@ test('link a {{pascal_case from_referenceable.name}} to a {{pascal_case to_refer
 {{#if (eq from_referenceable.hash_type "AgentPubKey")}}
     const baseAddress = alice.agentPubKey;
 {{else}}
+  {{#if (ne from_referenceable.hash_type "AnyLinkableHash")}}
     const baseRecord = await create{{pascal_case from_referenceable.name}}(alice.cells[0]);
+  {{/if}}
   {{#if (eq from_referenceable.hash_type "EntryHash")}}
     const baseAddress = (baseRecord.signed_action.hashed.content as NewEntryAction).entry_hash;
   {{else}}
+    {{#if (eq from_referenceable.hash_type "AnyLinkableHash")}}
```
matthme, line 57: There are still references to `AnyLinkableHash` here that I think need to be replaced with `ExternalHash` now if I'm not mistaken

c12i: Good catch, I switched to neovim recently, looks like nvim spectre search and replace failed here

matthme (approved):
I think it looks good to me know. I did notice some things that I think should be addressed at some point but they can be addressed independently of this PR:
* if you create a link type, the file name and function names have the target link in plural (I think this is because it uses the same naming logic as for collections where an anchor points to multiple entries). For example `validate_create_link_book_to_pages` instead of `validate_create_link_book_to_pake`

* There is a general problem it seems that files can get overwritten in case of name collisions. This can happen most easily with meta-data-only links whose file names are single word names, e.g. `book.rs` if the link type is named `book`. So in this case it would overwrite an existing `book.rs` file from a scaffolded `book` entry-type. Or the other way around, whichever is scaffolded first.

Issue #53, opened 2022-12-06
feat: Add ability to go back when adding entries


zippy:
I had the experience of realizing I made a mistake while adding a field, and had to ctrl-c out and start over.   It would be really nice to be able to navigate through the fields you have created, maybe with a list of the created fields so far, or something like that.

guillemcordoba:
Would it be sufficient to be able to remove fields from the list, so that you only have to re-do one field?

PR #383, opened 2024-10-09
feat: Improve `hc-scaffold entry-type` developer experience

c12i:
This PR introduces the following changes:

- Improves enum field_type selection, the field_name is assumed to be the label of the enum to avoid unnecessary prompts (closes https://github.com/holochain/Components/issues/68)
- Updates how we prompt users for field types, rather than asking them for the type first, we now ask them for the name of the field then the type.
- Add the option that allows the user to re-start building of their entry-type through a recursive call to `choose_fields`, incase there was an error or mistake when creating a field for their entry type. This is a first step towards the replay functionality, further improvements to this can be made with (a) separate PR(s) i.e to modify/add/delete specific fields rather than starting over (first of https://github.com/holochain/Components/issues/76 and https://github.com/holochain/scaffolding/issues/53)
- Refactor the `input_with_custom_validation` function to continue to re-prompt the user for input incase the validation fails rather than exiting the operation altogether (first step towards addressing https://github.com/holochain/scaffolding/issues/387)
- Greatly refactor the `FieldDefinition` and `EntryTypeReference` parsing logic to enhance resilience and better error handling

selection; refactor FieldType parsing

### DIFF
--- a/src/scaffold/entry_type/definitions.rs
+++ b/src/scaffold/entry_type/definitions.rs
@@ -234,7 +234,9 @@ impl FromStr for FieldDefinition {
     fn from_str(fields_str: &str) -> Result<Self, Self::Err> {
         let mut str_path = fields_str.split(':');
 
-        let field_name = str_path.next().context(format!("field_name is missing from: {}", fields_str))?;
+        let field_name = str_path
+            .next()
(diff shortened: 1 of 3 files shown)


matthme (approved):
Looks good :+1:

PR #398, opened 2024-10-22
chore: Update holochain dependencies

c12i:
This PR cleans up holochain dependency usage in the following ways:
- Removes unused `holochain_utils` crate; this dependency was only used with a single error variant which was not converted from anywhere in the code
- Replaces `test_utils` feature from holochain with `hdk`.


matthme (approved):
Looks like the hdk feature discussion has been resolved, so no blockers from my side anymore :)

PR #407, opened 2024-10-29
chore: update ui dependencies

### LINKED ISSUE #399: chore: Upgrade generated UI core dependencies
The following core dependencies need to be upgraded with any potential breaking changes addressed:

- [x] `@holochain-playground/cli`
- [x] `vite` and `vitest`
- [x] `react`/ `react-dom` et al.
- [x] `svelte` et al.
- [x] `vue` et al.
- [x] `lit` et al.

### LINKED ISSUE #408: [FEATURE] Bump lit version in lit template
Bump the lit template to lit version 3.x. It is currently using lit 2.6.

This would mean replacing:
```
    "@lit-labs/context": "^0.2.0",
    "@lit-labs/task": "^2.0.0",
    "lit": "^2.6.1"
```
with
```
    "lit": "^3.1.0",
    "@lit/context": "^1.1.0",
    "@lit/task": "^1.0.0",
```
Although I think `"@lit/task": "^1.0.0"` is not even used anywhere so it can potentially get removed completely.

It may also require smaller code changes but nothing major I think.

c12i:
This PR additionally adds some spacing to cli command outputs

Closes #399
Closes #408


templates/ui-frameworks/lit/web-app/ui/package.json.hbs
```diff
@@ -8,67 +8,31 @@
     "build:holo": "VITE_APP_IS_HOLO=true vite build",
     "package:holo": "{{(package_manager_command package_manager "build:holo" null)}} && rimraf dist.zip && cd dist && bestzip ../dist.zip *",
   {{/if}}
-    "lint": "eslint --ext .ts,.html . --ignore-path .gitignore && prettier \"**/*.ts\" --check --ignore-path .gitignore",
```
matthme, line 11: What's the motivation for removing the "lint" step here? I don't think it's a problem but I'm curious...

templates/ui-frameworks/lit/entry-type/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{#if crud.update}}edit-{{kebab_case entry_type.name}}.ts{{¡if}}.hbs
```diff
@@ -12,7 +12,6 @@ import { decode } from '@msgpack/msgpack';
   {{/each}}
 {{/uniq_lines}}
 
-import { sharedStyles } from '../../shared-styles';
```
matthme, line 15: Why are the sharedStyles being removed here? I'm assuming this will affect for example the button appearance?

templates/ui-frameworks/lit/entry-type/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{kebab_case entry_type.name}}-detail.ts.hbs
```diff
@@ -17,7 +17,6 @@ import { decode } from '@msgpack/msgpack';
 import './edit-{{kebab_case entry_type.name}}';
 {{/if}}
 
-import { sharedStyles } from '../../shared-styles';
```
matthme, line 20: Same here. It looks like this component uses the `alert` class for example which is defined in the shared styles so I would expect that this import should remain here.

templates/ui-frameworks/lit/web-app/ui/package.json.hbs
c12i: Yes, this was controversial. I noticed that eslint is only set up by default in the lit template. And I only removed it just to keep things simple and consistent with other templates. Perhaps I could keep it or add it to all templates with a set of similar starting rules in a `.eslintrc.js.hbs` template that can be re-used across the templates perhaps. This is something I can do in a separate PR hence why I removed it here for now. What do you think?

matthme: Okay :+1: I'm fine with removing it, I have no attachment to it.

templates/ui-frameworks/lit/entry-type/ui/src/{{dna_role_name}}/{{coordinator_zome_manifest.name}}/{{#if crud.update}}edit-{{kebab_case entry_type.name}}.ts{{¡if}}.hbs
c12i: I realized I could avoid this repetition by loading the styles to the `index.html` file

https://github.com/holochain/scaffolding/blob/5fc09c3d3ae34df0fe39508ee3b15fc62970dfd6/templates/ui-frameworks/lit/web-app/ui/index.html.hbs#L12

c12i: Just reverted the change though, it looks like loading the css from the html doesn't work as I expected

matthme: Yes I think it's required inside the component itself because lit uses shadow root which ignores global styles.

matthme (approved):
looks good to me know :+1:

PR #411, opened 2024-11-05
fix: Fix misplaced doc comments

### LINKED ISSUE #406: Many clippy warnings in scaffolded dnas
It looks like most are related to using `///` for expression comments instead of `//`

### LINKED ISSUE #413: [BUG] Broken link in generated `validate` function comments
## Describe the bug

The `validate` function contains a broken `docs.rs` URL

```
// You can read more about validation here: h// ocs.rs/hdi/latest/hdi/index.html#data-validation
```

## Expected behavior

The link should be `https://docs.rs/hdi/lates/hdi/index.html#data-validation`

## System information

-   OS: [e.g. Ubuntu 22.04, macOS 13.1]
-   Browser: [e.g. Chrome 120.0, Safari 16.3]
-   Scaffolding and Nix Version: [output from `hc-scaffold --version` or `hc-introspect` from the nix shell]

## Steps to reproduce

Scaffold Web App or Example and navigate to `validate` function in the `integrity` crate's `lib.rs` 

## Additional context

The bug is as a result of replacement of all instances of `//` with `///` and back to `//` which was a measure added with #382. The workaround might need to be re-worked as it can potentially lead to broken URLs inside comments in general

c12i:
Fixes #406
Fixes #413 (ensures `//` in URLs are not inadvertently replaced)

This PR also contains some additional code refactors to how the validate function codegen is handled

### DIFF
--- a/src/scaffold/zome/coordinator.rs
+++ b/src/scaffold/zome/coordinator.rs
@@ -72,9 +72,9 @@ pub fn initial_lib_rs(dependencies: Option<&Vec<String>>) -> TokenStream {
         pub enum Signal {}
 
         /// Whenever an action is committed, we emit a signal to the UI elements to reactively update them
-        /// Don't modify the for loop if you want the scaffolding tool to generate appropriate signals for your entries and links
         #[hdk_extern(infallible)]
         pub fn post_commit(committed_actions: Vec<SignedActionHashed>) {
+            /// Don't modify the for loop if you want the scaffolding tool to generate appropriate signals for your entries and links
             for action in committed_actions {
                 if let Err(err) = signal_action(action) {
                     error!("Error signaling new action: {:?}", err);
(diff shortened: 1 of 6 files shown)


src/scaffold/zome/coordinator.rs
```diff
@@ -72,9 +72,9 @@ pub fn initial_lib_rs(dependencies: Option<&Vec<String>>) -> TokenStream {
         pub enum Signal {}
 
         /// Whenever an action is committed, we emit a signal to the UI elements to reactively update them
+        /// Don't modify the for loop if you want the scaffolding tool to generate appropriate signals for your entries and links
         #[hdk_extern(infallible)]
         pub fn post_commit(committed_actions: Vec<SignedActionHashed>) {
-            /// Don't modify this loop if you want the scaffolding tool to generate appropriate signals for your entries and links
```
matthme, line 77: Was this comment moved due to clippy? It seems more likely to me that people would actually read this if the comment is right above the for loop. What about converting it into a double-shlash "normal" `//` comment instead of a doc comment? Or does it get removed then when parsing?

matthme (changes requested):

src/scaffold/zome/coordinator.rs
c12i: It was getting removed, but I figured out the underlying root cause, fixed now

matthme (approved):

c12i:
CI issue to be addressed when a newer version of tryorama is released

PR #414, opened 2024-11-11
feat: further improve ci runtime

### LINKED ISSUE #96: CI runtime improvements
* #93 proposes to use the nix built `hc-scaffold` for the integration tests.

* #95 proposes running `cargo test`. instead of `cargo test` we could add a [crane](https://github.com/ipetkov/crane)-based test derivation so the dependencies could be shared between the tests and the CLI. this assume we can agree on using a common cargo profile. any objections to using "release" for the test derivation?

* the script `run_tests.sh` could be parallelized using a matrix on github actions.

### LINKED ISSUE #412: fix: Further improve CI runtime performance
The CI pipeline is experiencing significant slowdown when using Holonix, specifically:

- Initial environment setup takes considerable time
- Repeated operations across different jobs may be redundant
- Cache utilization might not be optimal

c12i:
TODOS:

- [x] Remove holochain dependency
- [x] Configure cachix
- [x] Configure rust cache
- [x] Cancel in progress CI runs on pushing new refs
- [x] Optimize cranelib nix package

This PR looks to close #412 and will also close #96

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,6 +21,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: 1.80.0
+      - uses: Swatinem/rust-cache@v2
       - run: cargo test --no-fail-fast
 
   rustfmt:
@@ -43,6 +44,7 @@ jobs:
         with:
           toolchain: 1.80.0
           components: clippy
+      - uses: Swatinem/rust-cache@v2
       - run: cargo clippy -- -D warnings
 
   testbuild:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,6 +32,8 @@ jobs:
         with:
           name: holochain-ci
 
+      - uses: Swatinem/rust-cache@v2
+
       - name: rustfmt check
         run: |
           cd $GITHUB_WORKSPACE
@@ -73,6 +75,8 @@ jobs:
         with:
           name: holochain-ci
 
+      - uses: Swatinem/rust-cache@v2
+
       - name: Build and test
         run: |
           cd $GITHUB_WORKSPACE
@@ -105,6 +109,8 @@ jobs:
       - uses: cachix/cachix-action@v15
         with:
           name: holochain-ci
+
+      - uses: Swatinem/rust-cache@v2
 
       - name: Build and test
         run: |
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,6 +8,10 @@ on:
   pull_request:
     branches: [ develop, develop-0.1, develop-0.2 ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
 jobs:
   test_fmt_lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,9 +4,9 @@ on:
   # Trigger the workflow on push or pull request,
   # but only for the main branch
   push:
-    branches: [ develop, develop-0.1, develop-0.2 ]
+    branches: [develop, develop-0.1, develop-0.2]
   pull_request:
-    branches: [ develop, develop-0.1, develop-0.2 ]
+    branches: [develop, develop-0.1, develop-0.2]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yaml
```diff
@@ -1,77 +1,78 @@
...
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          install_url: https://releases.nixos.org/nix/nix-2.24.4/install
+          toolchain: 1.80.0
+          components: rustfmt
+      - run: cargo fmt --all -- --check
 
-      - uses: cachix/cachix-action@v15
+  clippy:
```
ThetaSinner, line 38: why separate these into different jobs? I'd do all Cargo work in one job as different steps. Otherwise you're cloning the repo and setting up the toolchain multiple times

.github/workflows/test.yaml
```diff
@@ -1,77 +1,78 @@
...
       - name: Install nix
-        uses: cachix/install-nix-action@v27
+        uses: cachix/install-nix-action@v30
         with:
           install_url: https://releases.nixos.org/nix/nix-2.24.4/install
 
       - uses: cachix/cachix-action@v15
         with:
-          name: holochain-ci
+          name: holochain-scaffolding-cli
```
ThetaSinner, line 74: No, you shouldn't try to cache everything. Set `extraPullNames` to `holochain-ci`

flake.nix
```diff
@@ -114,4 +126,10 @@
           };
         };
       };
+
+
+  nixConfig = {
```
ThetaSinner, line 131: Same here, should be both caches

src/versions.rs
```diff
@@ -10,11 +10,14 @@ pub const HC_SPIN_VERSION: &str = "^0.400.0-dev.3";
 /// npm: <https://www.npmjs.com/package/@holo-host/web-sdk>
 pub const WEB_SDK_VERSION: &str = "^0.6.20-prerelease";
 
-pub const HDI_VERSION: &str = holochain::HDI_VERSION;
+/// crates.io <https://crates.io/crates/hdi/versions>
+pub const HDI_VERSION: &str = "0.6.0-dev.3";
```
ThetaSinner, line 14: Does any documentation about how to bump the scaffolding to a new Holochain version need to change?

.github/workflows/test.yaml
```diff
@@ -1,77 +1,78 @@
...
+          components: clippy
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo clippy -- -D warnings
 
   testbuild:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        template: [ svelte, vue, lit, react, vanilla ]
+        template: [svelte, vue, lit, react, vanilla]
```
ThetaSinner, line 54: These will still run in parallel and rebuild the scaffolding a bunch of times, so the recommendation to build it once for Nix isn't done yet, is that for a follow up PR?

.github/workflows/test.yaml
ThetaSinner: This is actually wrong on Wind Tunnel. I will correct that next time I'm in there.

.github/workflows/test.yaml
c12i: I think separating them is good for failure isolation, faster feedback and parallelization. The time saved from repo cloning and toolchain setup is quite minimal, especially with caching set up. What do you think?

.github/workflows/test.yaml
c12i: Hmmm hadn't considered this. I think I can do this separately though

src/versions.rs
c12i: Good catch, updated the maintenance docs

flake.nix
c12i: Are you referring to extending the `substituters` and `trusted-public-keys` list here?

https://github.com/holochain/wind-tunnel/blob/main/flake.nix#L133

.github/workflows/test.yaml
ThetaSinner: I'd probably make these steps under one job, but if you've looked at the run times and you're happy then this works fine

flake.nix
ThetaSinner: Yes, although if you don't override the whole list, you don't need to add the NixOS cache. I'll probably change that again on Wind Tunnel. Just the Holochain CI

Yes it'd be the same values here, it's just the URL and a public key

.github/workflows/test.yaml
ThetaSinner: As in, a problem to solve on another PR?

c12i: Yes, I will need to weigh the impact of this. Probably will need to run the jobs sequentially so that only the first job builds scaffolding once. Running a job where there was a cache miss usually will take 35 minutes, then subsequent jobs using the cache from what I see take <= 10 minutes. So overall the jobs will succeed after (35min + <=10min * n) where the n is the number of jobs, this is much longer that running them in parallel which would take 35 minutes, much faster but not optimal. Anyway, yes, I will do this with a follow up PR, will keep the issue open.

c12i: Just pushed this commit though, I can keep it as part of this PR if this actually solves the problem

https://github.com/holochain/scaffolding/pull/414/commits/6853399e52585320b4e342cd6951d61004dc109c

ThetaSinner (approved):

PR #415, opened 2024-11-13
fix: Skip writing scaffold config for nixified custom templates

### LINKED ISSUE #247: Custom template nix wrapper results in `Error: unmatching template`
Using the [holochain-open-dev template](https://github.com/holochain-open-dev/template) with the recently merged https://github.com/holochain/scaffolding/pull/229 results in bad behavior between MacOs and linux users. This is because the template path gets written in the `package.json` like this: 

```json
  "hcScaffold": {
    "template": "/nix/store/c9lv5djdjlxxhk90g0idp4acapq2s0sv-hc-scaffold/template"
  }
```

And nix store paths are not consistent across platforms, so if I upload this repository and someone with another platform runs it, they are going to get an `Error: unmatching template`.

The custom template is internally setting the `--template` argument in every call to the scaffolding tool, and all that information is already embedded in the `hc-scaffold` binary that's available on the `nix develop` shell that the developers use. So in this case **I would actually like to disable that the template used to scaffold the project is written to the package.json**. Something like a `--disable-template-persistence` would suffice for me, but there is many different ways to implement something like this so I'm open to suggestions.

c12i:
~This PR introduces a global `--skip-config-check` flag that does as the name suggests, to provide an option to skip reading from `hcScaffold` config in the root package.json file. Using this flag will be mostly beneficial to users who scaffolded their hApps with a nix wrapper based custom template.~

This PR skips writing scaffold configs for web-apps or examples scaffolded from nixified custom templates

Closes #247

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,4 +50,9 @@ itertools = "0.13.0"
 colored = "2.1.0"
 dprint-plugin-typescript = "0.91.1"
 markup_fmt = "0.10.0"
-git2 = { version = "0.19.0", default-features = false, features = ["https", "ssh_key_from_memory", "vendored-libgit2", "vendored-openssl"] }
+git2 = { version = "0.19.0", default-features = false, features = [
+  "https",
+  "ssh_key_from_memory",
+  "vendored-libgit2",
+  "vendored-openssl",
+] }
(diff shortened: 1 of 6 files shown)


guillemcordoba (commented):
Ideally, I would want the scaffolding tool to not write the "hcSaffold" config option in the package.json, which I think it still does with this PR right?

guillemcordoba:
Awesome!

PR #419, opened 2024-11-21
fix: Remove `cap_secret` param from callZome api call

c12i:


jost-s (commented):
Wait, no, this is a mistake. cap_secret should not have been removed. I need to fix this in the JS client.

c12i:
Ahh I see, assumed this was intentional, will keep this as draft PR then

jost-s (approved):
LGTM

PR #422, opened 2024-12-03
chore: Bump holochain 0.3.6

matthme:


src/versions.rs
```diff
@@ -12,13 +12,13 @@ pub const HC_SPIN_VERSION: &str = "0.300.3";
 pub const WEB_SDK_VERSION: &str = "^0.6.19-prerelease";
 
 /// crates.io <https://crates.io/crates/hdi/versions>
-pub const HDI_VERSION: &str = "0.6.0-dev.3";
+pub const HDI_VERSION: &str = "0.4.6";
 
 /// crates.io <https://crates.io/crates/hdk/versions>
-pub const HDK_VERSION: &str = "0.5.0-dev.5";
```
matthme, line 18: Note that these 0.5.x numbers here have been accidentally introduced in a [recent] backport I think.

ThetaSinner: Oops, that's not good. I guess the tests probably worked because the HDK hasn't changed enough for that to fail. That's unfortunate :)

ThetaSinner (approved):

PR #425, opened 2024-12-04
Fix race condition in Svelte template

matthme:
Fixes a race condition where Svelte components would try to access the client context when it's not yet defined.


pdaoust:
Nice; I like the addition of a typedef for the client context. Kinda annoying that you had to edit all the files; thanks for doing that work!

(I looked over it for correctness but didn't scour to see if you'd missed any templates or example files; assuming you've done that ;) )

matthme:
I tried to catch all yes :) With search I think it's quite easy to see all occurrences so I'm quite confident.

pdaoust (approved):

PR #430, opened 2024-12-06
chore: Revert update to vite-plugin-svelte

pdaoust:
This reverts commit 981b106a1e5c7bcf46a43d128645a256d6074d68. Seems to introduce a packaging conflict.


pdaoust:
hang on, I don't yet know if this is an issue in 0.5

PR #410, opened 2024-10-31
chore: backport and bump 0.400.0

c12i:
## Requirements

- [x] Backport tagged PRs
- [x] Bump holochain
- [x] Bump hApp dependencies
   - [x] client-js
   - [x] tryorama (depends on https://github.com/holochain/tryorama/pull/245)
   - [x] hdi
   - [x] hdk

PR #432, opened 2024-12-10
chore: add backports to 0.4000.0 rc.3

c12i:
A mirror of #410.

Significant backports include:
- #407: ui dependency version bumps that should address issues faced while bumping the `vite-plugin-svelte` dependency by #431
- #414: apply ci runtime improvements from weekly which also include removal of `holochain` as a scaffolding dependency

Additional fixe
- [Pin typescript version for vue-tsc compatibility](https://github.com/holochain/scaffolding/pull/432/commits/8e6cfac06bc8e6ed2d86415cdfb85ec5f1c210e8)


matthme (approved):
I'm getting better at speed-testing scaffolding ;)
Looks good from the testing that I've done with the 4 different UI frameworks. And I did not spot something obviously wrong in the code changes.

src/lib.rs
```diff
@@ -97,10 +97,19 @@
...
+//!
+//!     ```bash
+//!     nix run github:holochain/holonix#hc-scaffold -- template new
+//!     ```
+//!     - Outside holonix if the cli was installed via `cargo install holochain_scaffolding_cli`
+//!
+//!     ```bash
+//!     hc-scaffold template new
+//!     ```
+//! 2. A new dir will be created in the current directory. Check this new folder in a version control system like git.
```
pdaoust, line 110:
```suggestion
//! 2. A new dir will be created in the current directory. Check this new folder into a version control system like git.
```

src/lib.rs
```diff
@@ -97,10 +97,19 @@
 //! To create a custom template, execute these steps:
 //!
 //! 1. Run this command:
```
pdaoust, line 99: This is such good info to have. One thing that I don't understand is what does `hc scaffold template clone` do? Would it be worth mentioning it here?

src/scaffold/entry_type/fields.rs
```diff
@@ -45,25 +46,99 @@ pub fn choose_fields(
...
+            match action {
+                0 => {
+                    // Change field
+                    if !fields.is_empty() {
+                        let field_to_change = Select::with_theme(&ColorfulTheme::default())
+                            .with_prompt("Select field to change")
+                            .items(
+                                &fields
+                                    .iter()
+                                    .map(|f| format!("{}: {}", f.field_name, f.field_type).italic())
```
pdaoust, line 80:
```suggestion
                                    .map(|f| format!("{}: {}", f.field_name, f.field_type))
```

The formatting on these menu options made me think it was informational text rather than things I could select, given that it was the only italic menu and italic had previously been used as info text. It took me a while to notice the menu!

I realise I'm giving this input on something that's only a backport PR though, so I could do this as a separate PR against 0.5 and 0.4.

pdaoust (changes requested):
I'm not going to let most my suggested changes be blockers; fine to go ahead with this as-is. I could create PRs for the suggestions later.

The only change I want to block on is that there seems to be a regression with Clap. When I'm trying to toggle 'checkboxes', it's hard to see the difference between selected and unselected:

image

We used to have this problem and it went away, but now it seems to be back again vs the released 0.4000.3:

image

Here the unselected check is so dark I can hardly see it, which makes it much easier to see what has been selected.

pdaoust (approved):

guides/maintenance.md
```diff
@@ -8,10 +8,10 @@ New development and bug fixes should target the `develop` branch. From there, ch
 
 There need to be some exceptions to this workflow to allow changes on maintenance branches that don't make sense to make on `develop`:
 
-- Bug fixes that are specific to a particular version of Holochain. This should be rare, but if Holochain were to introduce an API change between 0.1.x and 0.1.x+1 but NOT make that change for Holochain 0.2.x then the PR for Scaffolding would target `develop-0.1` directly.
-- Version bumps for `flake.nix`, `flake.lock` and `Cargo.lock` files should target the maintenance branches directly.
-- Version bumps for dependencies of Scaffolding in the `Cargo.toml`. For dependencies that are still on the same version as on `develop` it would be valid to target `develop` and back-port but you'll end up with a tricky lock file merge anyway.
-- Version bumps to Scaffolding itself don't make sense on `develop` so they should be made directly on the maintenance branches.
+-   Bug fixes that are specific to a particular version of Holochain. This should be rare, but if Holochain were to introduce an API change between 0.1.x and 0.1.x+1 but NOT make that change for Holochain 0.2.x then the PR for Scaffolding would target `develop-0.1` directly.
```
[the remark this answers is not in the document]
c12i: This is my editor that did the formatting for me. Can fix this with a separate PR since this is an upstream change from `develop`

src/cli/custom-template/README.md
```diff
@@ -41,24 +39,24 @@ Custom template for the [scaffolding tool](https://github.com/holochain/scaffold
...
             };
           };
       };
-}  
+}
 ```
 
 ---
 
 After this set up, you will be able to `nix develop` from inside your repository, and use the scaffolding tool as normal:
```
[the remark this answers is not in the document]
c12i: Thanks for catching this too, will fix upstream :)

src/lib.rs
c12i: `hc scaffold template clone` will create a template from the current template you are using, pretty meta but this was how it was originally defined

src/scaffold/entry_type/fields.rs
c12i: Ahh wasn't aware of this confusion, can look into this, thanks

PR #433, opened 2024-12-11
feat: add u8 field type support

### LINKED ISSUE #405: Add u8 & Vec<u8> to field types
AC:
- support `u8` fields
- support `Vec<u8>` field

c12i:
This PR will close #405.

Additionally, the representation of `Vec<u8>` field types in UI code is a `Uint8Array` instead of `Array<number>`.


matthme:
I'm noticing, at least in the Vue framework so far, that a `u8` field does not offer the option to scaffold a UI widget, while `u32` does. I think `u8` could use that same widget?

matthme (changes requested):
Left one comment regarding adding the slider component as well for `u8`. Not sure anyone will ever use a `u8` type outside of a vector though but if it's not too much effort it would be nice I think to have the same behavior as `u32` in this regard.

templates/ui-frameworks/vue/field-types/u8/Slider/edit/render.hbs
```diff
@@ -0,0 +1,2 @@
+<label for="{{label}}">{{label}}</label>
+<input name=""{{label}} type="range" min="0" max="255" v-model="{{variable_to_read}}" />
```
matthme, line 2:
```suggestion
<input name="{{label}}" type="range" min="0" max="255" v-model="{{variable_to_read}}" />
```

templates/ui-frameworks/svelte/field-types/u8/Slider/edit/render.hbs
```diff
@@ -0,0 +1,2 @@
+<label for="{{label}}">{{label}}</label>
+<input name=""{{label}} type="range" min="0" max="255" bind:value="{{variable_to_read}}" />
```
matthme, line 2:
```suggestion
<input name="{{label}}" type="range" min="0" max="255" bind:value={ {{variable_to_read}} } /
```

matthme: I saw that there are more of these in other svelte slider components

matthme:
Now it offers to use a Slider also when a `Vec<u8>` is scaffoldded. I'm not sure that's expected behavior. And also, when I scaffold an entry type with a `u8` field and `Vec<u8>` field, the type of the `Vec<u8>` field in the Create and Edit Svelte UI components is scaffolded as `Array<number>` instead of `Uint8Array` although the type in the `types.ts` file is correctly a `Uint8Array`.

templates/ui-frameworks/vue/field-types/u8/Slider/edit/render.hbs
c12i: Looks my editor did a little bit of bad formatting, thanks for catching this

c12i: Turns out it wasn't an autoformat issue afterall, it seems like it was an issue with other existing `vue` and `svelte` field types, this must have went unnoticed in previous code reviews

templates/ui-frameworks/svelte/field-types/u8/Slider/edit/render.hbs
c12i: Good catch, another one that went unnoticed previously

matthme (approved):
Looks good to me now :+1:

PR #439, opened 2024-12-23
fix: gracefully handle number prefixed names

### LINKED ISSUE #435: [BUG] cli crashes when using numeric prefixed names
## Description

When using `hc scaffold web-app` with a name that starts with numbers and contains underscores (e.g., 03_to_04), the scaffolding process crashes during DNA generation and build. Similar behaviour can also be reproduced when scaffolding an `entry-type` also containing a numeric prefixed names.

## Expected behavior

Successful scaffolding of a web app or entry-type

c12i:
Fixes #435

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -149,7 +149,7 @@ pub fn input_no_whitespace(prompt: &str) -> ScaffoldResult<String> {
         .with_prompt(prompt)
         .interact_text()?;
 
-    while let Err(e) = check_no_whitespace(&input, "Input") {
+    while let Err(e) = validate_input(&input, "Input") {
         println!("{}", e.to_string().red());
         input = Input::with_theme(&ColorfulTheme::default())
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -167,17 +167,27 @@ pub fn check_case(input: &str, identifier: &str, case: Case) -> ScaffoldResult<(
             "{identifier} must be {case:?} Case",
         )));
     }
+    if input.chars().next().map_or(false, char::is_numeric) {
+        return Err(ScaffoldError::InvalidStringFormat(format!(
+            "{identifier} must not start with a numeric"
```
ThetaSinner, line 172:
```suggestion
            "{identifier} must not start with a number"
```

Or maybe "numeric value"?

ThetaSinner (approved):
Nice, thanks for fixing this :)

PR #441, opened 2025-01-02
Update locks and nixpkgs

ThetaSinner:


c12i (commented):
Thanks 👍🏽

c12i (approved):

ThetaSinner:
This doesn't look like something I've broken?

ThetaSinner:
No.. that won't fix it. The Rust version is pinned in Holonix deliberately. We don't update the Rust version on maintenance versions as a general rule. This is the actual generated hApp that has problems

c12i:
Hmm yes, I remember that 0.3 is on 1.78. Let me have a look at this further locally then

ThetaSinner:
I think upgrading wasmer on 0.3 is too big a risk. I'm tempted to suggest we pin `home`. Trying that out

ThetaSinner:
That seems to solve the problem locally, let's see if the CI agrees :)

src/scaffold/zome.rs
```diff
@@ -345,6 +345,7 @@ pub fn add_common_zome_dependencies_to_workspace_cargo(
         &format!("={}", versions::HDK_VERSION),
     )?;
     let file_tree = add_workspace_external_dependency(file_tree, "serde", "1.0")?;
+    let file_tree = add_workspace_external_dependency(file_tree, "home", "=0.5.9")?;
```
ThetaSinner, line 348: I can't add a removal comment here, people will have to figure this one out

c12i: Yea, this is handled programatically, maybe a top level comment in the workspace cargo toml here https://github.com/holochain/scaffolding/blob/ce3f9daee5709f103b28033e591b1bf3a242706c/src/scaffold/app/cargo.rs#L12

ThetaSinner: Doesn't seem to work. I'm thinking the TOML load/write is not preserving comments

c12i: Ah shucks, I think the comments in the workspace crates should be a good indicator to remove the workspace dependency when upgrading, we can leave this as is for now

PR #444, opened 2025-01-07
chore: bump weekly `0.500.0 dev.5`

c12i:


ThetaSinner (approved):
Any reason not to bump to dev.13? That released yesterday

PR #462, opened 2025-02-10
chore(vue): Migrate from options API to composition API

### LINKED ISSUE #459: [CHORE] Migrate Vue templates to vue composition API
We are currently running Vue3 with the options API. This issue recommends migrating to the composition API which is much easier to develop with

c12i:
Closes #459 

This PR migrates from the options API to the composition API

### DIFF
--- a/templates/ui-frameworks/vue/web-app/ui/src/ClientProvider.vue.hbs
+++ b/templates/ui-frameworks/vue/web-app/ui/src/ClientProvider.vue.hbs
@@ -1,0 +1,13 @@
+<template>
+  <progress v-if="loading"></progress>
+  <div v-else-if="error" class="alert">
+    Error connecting to Holochain: {{{{raw}}}} {{error.message}}.{{{{/raw}}}}
+  </div>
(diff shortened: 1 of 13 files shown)


templates/ui-frameworks/vue/web-app/ui/src/ClientProvider.vue.hbs
```diff
@@ -0,0 +1,13 @@
+<template>
+  <progress v-if="loading"></progress>
```
c12i, line 2: This provider acts as the guard to prevent using the client before its loaded

pdaoust:
Just looking through the code to the best of my abilities (I don't know either framework very well) and it looks like you've moved a lot of the Svelte templates' complexity from various templates into the provider and context, which is great. Can you share a general overview of what's happening here so I can understand it more quickly? Like, what's the process from page load to a usable client? What are the error cases, and are they recoverable without a page refresh?

c12i:
Scope changed, closing in favor of #473

PR #467, opened 2025-02-18
Expand merge helper scopes

guillemcordoba:
This PR adds support to more scopes for the `merge` helper. Previously, only "{}" was valid, but now it has been expanded to "{}", "[]", "()" and "<>". 

I encountered the need for this in a downstream project and thought it would be good to have it upstream. Let me know if I should rebase the PR to another branch, although I would need this merged in to 0.4 main.


c12i:
@guillemcordoba Just changed the base branch. We recently moved to using `main-*` prefixed branches over `develop-*` and the older branches will be deleted soon

c12i (approved):

PR #474, opened 2025-03-19
chore(ci): Disable fail-fast ci job strategy option

c12i:
Disabling to allow template tests to continue to run when one fails.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -81,6 +81,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [nix_build]
     strategy:
+      fail-fast: false
       matrix:
         template: [svelte, vue, lit, react, vanilla]
     steps:


ThetaSinner:
Why not get feedback about which things are passing just because one is broken? Doesn't that mean more CI runs when more than one thing is broken?

It'd be useful to have the justification for why this is being changed in the PR description

c12i:
@ThetaSinner added the description my bad. The justification behind this is to prevent tryorama tests that fail for one template to not stop jobs for other templates, particularly now due to sporadic flaky tests that would need re-running. What do you think?

mattyg (approved):

mattyg:
> Why not get feedback about which things are passing just because one is broken? Doesn't that mean more CI runs when more than one thing is broken?

I think you're describing `fail-fast: true`, the previous behavior. This PR changes it to `fail-fast: false`, so CI will continue to run if a matrix job fails.

ThetaSinner (approved):

PR #479, opened 2025-03-20
build: bump nixpkgs for crane support

mattyg:
I was getting this warning when entering the nix shell:

```trace: warning: crane requires at least nixpkgs-24.11, supplied nixpkgs-24.05```

This PR bumps nixpkgs to 24.11

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -137,16 +137,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1717179513,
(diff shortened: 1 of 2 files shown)


ThetaSinner (approved):

c12i (approved):
Thanks

mattyg:
Huh this test failure is a bit concerning -- its arising when alice deletes a link, then we `await dhtSync`, then bob gets the link. It seems like the dhts are not synced still. Looking back through past CI runs it looks like this has been occuring inconsistently for a while. 

I'm going to raise an issue in tryorama, and assume its an issue with the `dhtSync` function, not holochain.

PR #484, opened 2025-03-26
chore: backport to 0.4 line

c12i:


jost-s:
Are there no crates to be updated?

c12i:
> Are there no crates to be updated?

Not this time, this backport is only updating the ci workflow for `main-0.4`. Would you recommend a version bump of the crate regardless?

Issue #533, opened 2026-01-07
Switch from tryorama to sweettest


ThetaSinner:
Since the Holochain team are planning to support only one test framework, rather than both sweettest and tryorama, the scaffolding needs to switch. It currently generates tryorama tests. It should generate sweettest tests instead.

**Tasks:**
* Create new Rust sweettest templates (test files, common helpers, Cargo.toml)
  * call conductor methods directly instead of using a Holochain binary and `hc` and calling the conductor API                                                                                                                                                                                                                                           
* Update scaffolding tool source code (remove npm/Tryorama version constants and template data)
  * Update src/versions.rs (remove Tryorama/npm version constants)                                                                                                                                                                                                                                                    
  * Update src/templates/web_app.rs (remove npm template data)
  * Verify template rendering logic works with new .rs.hbs templates
* Delete all old TypeScript test templates

PR #536, opened 2026-01-16
refactor!: delete react, lit and vue as selectable ui templates

### LINKED ISSUE #531: Provide support for a single UI framework
The scaffolding tool provides templates for several UI frameworks. This requires extra maintenance for the Holochain team and adds less value that it used to work the availability of AI tools. If a user would prefer to switch to a new UI framework, it can be done using AI tooling. 

The proposal is to support only Svelte. It is a capable framework and should meet the needs of most users.

AC:
- [ ] It is still possible to scaffold a Svelte app, and this is tested on CI.
- [ ] Support for other UI frameworks such as React, Vue, and Lit is removed.

jost-s:
depends on #533 

resolves #531 

- `hello_world` example can only be scaffolded with `vanilla` template.
- `forum` can only be scaffolded with `svelte` template.


README.md
```diff
@@ -26,7 +26,8 @@ hc-scaffold --version
 
 ## Usage
 
-Refer to [the holochain developer instructions](https://developer.holochain.org/get-building/) to know how you can use the scaffolding tool to create your own apps.
+Refer to [the holochain developer instructions](https://developer.holochain.org/get-building/) to know how you can use
```
ThetaSinner:
```suggestion
Refer to [the holochain developer instructions](https://developer.holochain.org/get-building/) to learn how you can use
```

ThetaSinner (dismissed):
Great!

mattyg (commented):
I was assuming we would drop vanilla as well so there was only a single set of templates for generating a UI. Was that not the intention?

PR #562, opened 2026-05-08
fix: version mismatch in templates

### LINKED ISSUE #561: [BUG] Conductor config version mismatch
## Describe the bug

The CI is currently failing ([example](https://github.com/holochain/scaffolding/actions/runs/24791016064/job/72550908684?pr=560)), looking at the logs it seems that this repo is trying to use the Holochain conductor config from `holochain-0.7.0-dev.16` but it should be using `holochain_conductor_config v0.7.0-dev.20`.

Also, the `perl` package is missing from the `ci` `devShell` and maybe the `default` one too. It should be added as it is required to build OpenSSL.

AC:
- [x] The CI should pass
- [x] `perl` package is part of the devShells.

cdunster:
This PR is mainly to update the versions of the Holochain crates used by the templates, but it also fixes some format and linting issues, uses the same Rust toolchain version in all Nix derivations, and adds missing packages to Nix derivations.

Closes #561.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -98,6 +98,7 @@
             ]) ++ (with pkgs; [
               nodejs_24
               binaryen
+              perl
             ]) ++ [
               self'.packages.hc-scaffold
             ];
--- a/flake.nix
+++ b/flake.nix
@@ -37,81 +37,85 @@
               '';
         };
         systems = builtins.attrNames inputs.holonix.devShells;
-        perSystem = { inputs', self', system, pkgs, lib, ... }: {
-          formatter = pkgs.nixpkgs-fmt;
+        perSystem = { inputs', self', system, pkgs, lib, ... }:
+          let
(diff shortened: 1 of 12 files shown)


flake.nix
```diff
@@ -33,83 +34,88 @@
...
+                holochain
+                lair-keystore
+                hn-introspect
+              ]) ++ (with pkgs; [
+                nodejs_24
+                binaryen
+                perl
+              ]) ++ [
+                self'.packages.hc-scaffold
+                rustToolchain
```
cdunster, line 107: This uses the toolchain version from the `rust-toolchain.toml` file. We could also use the one from Holonix instead but we should probably use the same version everywhere instead of mixing them.

cdunster:
@coderabbitai review

ThetaSinner (approved):

PR #571, opened 2026-07-13
feat: add ELECTRON_BINARY env var override for flake.nix template

cdunster:
`ELECTRON_BINARY` is used by `hc-spin` to set the path to the Electron binary to use.

This forces `hc-spin` to use the Electron binary from nixpkgs, fixing known problems with running on NixOS and with installing Electron on certain Node.js versions.

### DIFF
--- a/src/cli/custom-template/flake.nix
+++ b/src/cli/custom-template/flake.nix
@@ -16,6 +16,8 @@
       devShells.default = pkgs.mkShell {
         inputsFrom = [ inputs'.holonix.devShells.default ];
 
+        ELECTRON_BINARY = "${pkgs.electron_42}/bin/electron"; # Force `hc-spin` to use Electron from nixpkgs.
+
         packages = (with pkgs; [ nodejs_24 binaryen ]);
 
         shellHook = ''
--- a/src/cli/custom-template/flake.nix
+++ b/src/cli/custom-template/flake.nix
@@ -18,7 +18,11 @@
 
         ELECTRON_BINARY = "${pkgs.electron_42}/bin/electron"; # Force `hc-spin` to use Electron from nixpkgs.
 
-        packages = (with pkgs; [ nodejs_24 binaryen ]);
+        packages = (with pkgs; [
+          nodejs_24
+          binaryen
+          perl # Required to build OpenSSL from source
+        ]);
 
         shellHook = ''
           export PS1='\[\033[1;34m\][holonix:\w]\$\[\033[0m\] '
--- a/src/scaffold/app/nix.rs
+++ b/src/scaffold/app/nix.rs
@@ -35,6 +35,7 @@ pub fn flake_nix() -> FileTree {
         packages = (with pkgs; [
           nodejs_24
           binaryen
+          perl # Required to build OpenSSL from source
         ]);
 
         shellHook = ''
--- a/src/scaffold/app/nix.rs
+++ b/src/scaffold/app/nix.rs
@@ -30,8 +30,10 @@ pub fn flake_nix() -> FileTree {
       devShells.default = pkgs.mkShell {
         inputsFrom = [ inputs'.holonix.devShells.default ];
 
+        ELECTRON_BINARY = "${pkgs.electron_42}/bin/electron"; # Force `hc-spin` to use Electron from nixpkgs.
+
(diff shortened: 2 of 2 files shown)


src/scaffold/app/nix.rs
```diff
@@ -30,9 +30,12 @@ pub fn flake_nix() -> FileTree {
       devShells.default = pkgs.mkShell {
         inputsFrom = [ inputs'.holonix.devShells.default ];
 
+        ELECTRON_BINARY = "${pkgs.electron_42}/bin/electron"; # Force `hc-spin` to use Electron from nixpkgs.
+
         packages = (with pkgs; [
-          nodejs_22
+          nodejs_24
```
cdunster, line 36: I can't imagine that there's a reason that this was an older version than everywhere else, but it's worth others confirming if they can. This is the actual version added to templated `flake.nix` files, so this is most likely the most important place to have it correct.

src/scaffold/zome/coordinator.rs
```diff
@@ -45,6 +45,9 @@ holochain_serialized_bytes = {{ workspace = true }}
 holochain = {{ workspace = true, features = ["wasmer_sys", "transport-iroh", "test_utils"] }}
 tokio = {{ workspace = true }}
 
+# Force build of OpenSSL from source
+openssl = {{ version = "0.10", features = ["vendored"] }}
```
cdunster: I'm not a big fan of this approach, and I think that it's only the `forum` template that requires it, but it works and is the only fix right now unless we fix the feature flags in holochain_metrics that include a non-vendorded version of OpenSSL.

veeso (dismissed):

src/scaffold/zome/coordinator.rs
jost-s: Why is openssl now required, when it doesn't seem to be broken up until now?

cdunster: It's always been broken on NixOS just for the forum template but it's just an edge case that I think no one except me has run into. I think that `holochain_metrics` expects the OpenSSL dev libraries to be installed locally which I think is a mistake that we can fix in that crate instead of here.

ThetaSinner: We've made a mistake somewhere if openssl has made its way into a coordinator build

ThetaSinner: Oh I see the thread below agrees, yes can we please fix this in the `holochain_metrics` crate then

cdunster: Fixed in `holochain_metrics` in `v0.6.3` and now this branch uses that version.

cdunster: Fixed in `holochain_metrics` in `v0.6.3` and now this branch uses that version.

jost-s (approved):

src/scaffold/app/nix.rs
```diff
@@ -30,9 +30,12 @@ pub fn flake_nix() -> FileTree {
...
           binaryen
+          perl # Required to build OpenSSL from source
```
jost-s, line 38: Wait, this shouldn't be needed anymore then?

src/cli/custom-template/flake.nix
```diff
@@ -16,7 +16,13 @@
       devShells.default = pkgs.mkShell {
         inputsFrom = [ inputs'.holonix.devShells.default ];
 
-        packages = (with pkgs; [ nodejs_24 binaryen ]);
+        ELECTRON_BINARY = "${pkgs.electron_42}/bin/electron"; # Force `hc-spin` to use Electron from nixpkgs.
+
+        packages = (with pkgs; [
+          nodejs_24
+          binaryen
+          perl # Required to build OpenSSL from source
```
jost-s, line 24: And this

ThetaSinner (approved):

src/scaffold/app/nix.rs
cdunster: This is for the scaffolded projects that require `perl` so that they can build OpenSSL from source, which is a requirement of some Holochain packages.

src/cli/custom-template/flake.nix
cdunster: Same reply as above.

PR #575, opened 2026-07-27
fix(link-type): compare link targets as any linkable hashes

### LINKED ISSUE #546: [BUG] Link from external hash to external hash doesn't compile
#### Steps to reproduce

```sh
$ hc-scaffold link-type
✔ Link from which entry type? · ExternalHash
✔ What name should be given to the link for this hash? · some
✔ Link to which entry type? · Agent
✔ Which role does this agent play in the relationship ? (eg. "creator", "invitee") · other
✔ Should the link be bidirectional? · yes
✔ Can the link be deleted? · yes

Link type scaffolded!

$ npm t

> test
> npm run build:happ && cargo test


> build:happ
> npm run build:zomes && hc app pack workdir --recursive


> build:zomes
> RUSTFLAGS='--cfg getrandom_backend="custom"' cargo build --release --target wasm32-unknown-unknown
...
error[E0308]: mismatched types
  --> dnas/dena/zomes/coordinator/zoma/src/some_to_others.rs:98:47
   |
98 |         if link.target.clone().into_hash() == input.base_some.clone().into_hash() {
   |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `HoloHash<AnyLinkable>`, found `HoloHash<External>`
   |
   = note: expected struct `HoloHash<AnyLinkable>`
              found struct `HoloHash<hdk::prelude::hash_type::External>`
help: call `Into::into` on this expression to convert `HoloHash<hdk::prelude::hash_type::External>` into `HoloHash<AnyLinkable>`
   |
98 |         if link.target.clone().into_hash() == input.base_some.clone().into_hash().into() {
   |                                                                                   +++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `zoma` (lib) due to 1 previous error
```

**Tasks:**
- Fix bug and simplify comparison:
```rust
if link.target == input.base_some.clone().into() {
```

veeso:
## What changed

Link deletion now compares both addresses in the same format. This applies to both normal and bidirectional links. Regression coverage was added for deletable bidirectional external hash links.

## Why

Generated zomes could not compile when an external hash was used in a deletable link because the generated comparison used incompatible hash types.

## Notes

Validated with all 102 library tests, Clippy with warnings denied, and a fresh Nix application build through npm run build:happ. The later native cargo test stage is currently blocked by an unrelated incompatibility between the scaffolded Holochain development dependencies.

Closes #546

### DIFF
--- a/src/scaffold/link_type/coordinator.rs
+++ b/src/scaffold/link_type/coordinator.rs
@@ -521,12 +521,8 @@ fn remove_link_handlers(
     let pascal_link_type_name =
         format_ident!("{}", link_type_name(from_referenceable, to_referenceable));
 
-    let from_link_hash_type_code = hash_type_code_from_referenceable(to_referenceable);
-
     let bidirectional_remove = if bidirectional {
         {
-            let from_inverse_hash_type = hash_type_code_from_referenceable(from_referenceable);
-
             quote! {
                 let links = get_links(
                     LinkQuery::try_new(
@@ -535,7 +531,7 @@ fn remove_link_handlers(
                     GetStrategy::default(),
                 )?;
                 for link in links {
-                    if #from_inverse_hash_type == input.#base_field_name.clone().into_hash() {
+                    if link.target == input.#base_field_name.clone().into() {
                         delete_link(link.create_link_hash, GetOptions::default())?;
                     }
                 }
@@ -559,7 +555,7 @@ fn remove_link_handlers(
                 GetStrategy::default(),
             )?;
             for link in links {
-                if #from_link_hash_type_code == input.#target_field_name.clone().into_hash() {
+                if link.target == input.#target_field_name.clone().into() {
                     delete_link(link.create_link_hash, GetOptions::default())?;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/scaffold/link_type/coordinator.rs
```diff
@@ -521,12 +521,8 @@ fn remove_link_handlers(
     let pascal_link_type_name =
```
veeso, line 521: only the changes in this file are relevant for the fix.
All the other changes have been made to fix the ci.

src/scaffold/zome/coordinator.rs
```diff
@@ -30,20 +30,22 @@ name = "{zome_name}"
...
 [dev-dependencies]
-holochain = {{ workspace = true, features = ["wasmer-sys-cranelift", "transport-iroh", "test_utils"] }}
+holochain = {{ workspace = true, features = ["test_utils", "transport-iroh", "wasmer-sys-cranelift"] }}
 tokio = {{ workspace = true }}
+# Holochain dev.32 requires Wasmer 7.1; Wasmer 7.2 is not API-compatible.
+wasmer = {{ workspace = true }}
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+name = "{zome_name}"
```
ThetaSinner: Necessary to move this? I think of the sections as ordered by meaning and where you'd expect to look for them - rather than alphabetically. I'm fine with sorting within sections, like dependencies, that's helpful

src/scaffold/zome.rs
```diff
@@ -322,25 +322,27 @@ pub fn add_common_zome_dependencies_to_workspace_cargo(
...
         "version".to_string(),
-        toml::Value::String(HOLOCHAIN_VERSION.to_string()),
+        toml::Value::String(format!("={HOLOCHAIN_VERSION}")),
     );
     dep_holochain.insert("default-features".to_string(), toml::Value::Boolean(false));
     let file_tree =
         add_workspace_dependency(file_tree, "holochain", &toml::Value::Table(dep_holochain))?;
     let file_tree = add_workspace_external_dependency(file_tree, "tokio", "1.49")?;
+    let file_tree =
+        add_workspace_external_dependency(file_tree, "wasmer", &format!("={WASMER_VERSION}"))?;
```
ThetaSinner: This is because of the wasmer 0.7.2 release that removes features we use? I thought we had that pinned in Holochain? If so, we should be able to mark this as a temporary change I think?

src/scaffold/zome/coordinator.rs
veeso: reverted; I saw you solved build issues in another pr

src/scaffold/zome.rs
veeso: reverted; I saw you solved build issues in another pr

ThetaSinner (approved):

PR #576, opened 2026-07-27
Upgrade to Holochain 0.7.0-rc.4 (hdi 0.8.0-rc.3 / hdk 0.7.0-rc.3)

ThetaSinner:
## Summary
- Rewrites the `syn`/`quote`-based integrity zome `validate()` and coordinator `signal_action` codegen for the hdi API changes since 0.7.0-dev.23 (`FlatOp` variants renamed, `EntryCreationAction` replaced by `TypedAction<EntryCreationData>`, `Action` becoming a `{header, data}` struct, link validation functions collapsed to a single `TypedAction<CreateLinkData>` param), and drops a Cargo feature that no longer exists.
- Bumps `HOLOCHAIN_VERSION`/`HDI_VERSION`/`HDK_VERSION` to `0.7.0-rc.4`/`0.8.0-rc.3`/`0.7.0-rc.3` to route around a crates.io dependency drift where `holochain`'s own transitive deps had moved ahead of what `0.7.0-rc.3` itself was built against, breaking a fresh lockfile resolution for the sweettest dev-dependency.
- Adds `CLAUDE.md` documenting the upgrade process (where versions live, where the HDI/HDK API surface is hard-coded, how to find the target API shape, how to verify a change) for next time.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -160,7 +160,59 @@ os.write(master, b"\x1b[B\r")  # Down, Enter — e.g. to answer "No"
   features enabled at the workspace level, that's likely a hand-edit on top
   of what was scaffolded, not something to replicate by default.
 
-## 6. If something looks like an upstream regression
+## 6. "It compiled yesterday" — crates.io drift during active RC cycles
+
+During an active RC cycle, `holochain`'s own transitive deps (`holochain_state`,
+`holochain_cascade`, `holochain_conductor_api`, ...) can get republished to
+crates.io *ahead of* a coordinated `holochain` release, breaking a fresh
+`cargo generate-lockfile` even though nothing in this repo or in `versions.rs`
+changed. Symptom: `cargo check --workspace --all-targets` fails **inside the
+`holochain` crate's own compilation** (not in the generated zome code) with
+errors like "no field `X`" or "this method takes N arguments" pointing at
+paths under `~/.cargo/registry/.../holochain-X.Y.Z-rc.N/...`.
+
+Diagnose by checking what actually resolved:
+
+```sh
+grep -A1 '^name = "holochain'  Cargo.lock   # in the affected app/test workspace
+```
+
+If a sibling crate (e.g. `holochain_state`) resolved to a *newer* pre-release
+than `holochain` itself, that's the drift. Confirm by trying to pin it back:
+
+```sh
+cargo update -p holochain_state --precise <holochain's-own-version>
+```
+
+If that fails ("candidate versions found which didn't match"), the sibling
+has already moved on and there's no going back — `holochain`'s own Cargo.toml
+now requires the newer sibling.
+
+**Do NOT try to decouple the `holochain` dev-dependency's version from
+`hdi`/`hdk`** (e.g. "pin the zome to hdi=rc.2 but let sweettest use a newer
+holochain") — this is not resolvable by Cargo. Enabling `test_utils`/
+`sweettest` on `holochain` activates its own optional `hdk` dependency, and
+Cargo unifies dependency resolution workspace-wide, so the exact `hdk` pin
+used by the zome crates conflicts with whatever `holochain` wants
+transitively. It fails outright with "failed to select a version for `hdk`".
+
+**The actual fix is to bump `hdi`/`hdk`/`holochain` together** to the next
+tag where upstream re-coordinated. Before assuming that requires new codegen
+work, diff `crates/hdi/src` and `crates/hdk/src` between the two tags — if
+empty, the bump is a pure version-string change in `versions.rs`, since the
+breakage was entirely inside `holochain`'s own conductor code, not in the
+zome-facing API:
+
+```sh
+gh api repos/holochain/holochain/compare/hdi-OLD...hdi-NEW \
+  -q '.files[] | .filename' | grep -E '^crates/(hdi|hdk)/src/'
+```
+
+Re-run the full §4 verification loop after any such bump — don't assume it's
+safe just because the source diff was empty; confirm `cargo check --workspace
+--all-targets` actually passes.
+
+## 7. If something looks like an upstream regression
 
 If the target release seems to have dropped a helper or introduced a logic
 gap with no replacement (not just a rename), don't paper over it with a
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,168 @@
+# Upgrading Holochain versions
+
+This repo is a code generator: it writes Rust (hdi/hdk) source into scaffolded
+apps using `syn`/`quote` AST manipulation (not just Handlebars templates), so
+bumping the pinned Holochain version can require updating Rust code in this
+repo, not just a version string.
+
+## 1. Where the versions live
+
+`src/versions.rs` holds five constants:
+
+- `HOLOCHAIN_VERSION` (crates.io `holochain`)
+- `HDI_VERSION` (crates.io `hdi`)
+- `HDK_VERSION` (crates.io `hdk`)
+- `HOLOCHAIN_CLIENT_VERSION` (npm `@holochain/client`)
+- `HC_SPIN_VERSION` (npm `@holochain/hc-spin`)
+
+Find the matching set for a target Holochain release from the
+`holochain/holochain` GitHub repo tags (`holochain-X.Y.Z`, `hdi-X.Y.Z`,
+`hdk-X.Y.Z`) or crates.io/npmjs listings — the four crates are released
+together but with independent version numbers, so don't assume they move in
+lockstep.
+
+This repo does **not** pin `holochain_client` or `holochain_types` as Rust
+crates anywhere (only the npm client and the crates above). If a reference
+app has those in its workspace `Cargo.toml`, that's app-specific
+customization, not something this tool generates — don't add it here on
+their account.
+
+## 2. Where the HDI/HDK API surface is hard-coded
+
+The `validate()` callback and its match arms are built programmatically via
+`syn`/`quote`, so breaking changes to the `hdi::flat_op` module (`FlatOp`,
+`OpEntry`, `OpUpdate`, `OpDelete`, `OpLink`, `OpRecord`, `OpActivity`,
+`TypedAction`) or to the `Action`/`ActionData` shape need matching edits in:
+
+- `src/scaffold/zome/integrity.rs` — `initial_lib_rs()`: the base `validate()`
+  skeleton generated for a brand-new integrity zome (before any entry/link
+  types exist), plus the always-static `AgentActivity`/`CreateAgent` arm.
+- `src/scaffold/entry_type/integrity.rs` — `render_entry_definition_file()`
+  (the `validate_create_X`/`validate_update_X`/`validate_delete_X` function
+  signatures) and `add_entry_type_to_validation_arms()` + its `handle_*_arm`
+  helpers (the logic that finds the right match arm in an existing
+  `validate()` and inserts a new entry-type arm into it).
+- `src/scaffold/link_type/integrity.rs` — same idea for link types:
+  `validate_referenceable()`, the `validate_create_link_X`/
+  `validate_delete_link_X` signatures in `add_link_type_to_integrity_zome()`,
+  `add_link_type_to_validation_arms()`, and `add_link_type_signals()` /
+  `signal_action_match_arms()` (the coordinator-side `Signal::LinkCreated`/
+  `LinkDeleted` codegen).
+- `src/scaffold/entry_type/coordinator.rs` — `signal_action_match_arms()`
+  (the coordinator-side `Signal::EntryCreated`/`Updated`/`Deleted` codegen)
+  and the `match action.hashed.content...` skeleton it shares with the
+  file above.
+- `src/scaffold/zome/coordinator.rs` — `initial_cargo_toml()`: the
+  per-coordinator-zome dev-dependency feature list for the `holochain` crate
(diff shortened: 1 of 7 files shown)


src/scaffold/entry_type/integrity.rs
```diff
@@ -525,16 +525,24 @@ fn handle_store_record_arm(
...
-                    // Change empty invalid to match on entry_type
-                    *op_record_arm.body = syn::parse_quote! {match app_entry {}};
+                    // Change empty invalid to match on entry_type, after narrowing the action
+                    // down to a `TypedAction<EntryCreationData>`
+                    *op_record_arm.body = syn::parse_quote! {
+                        {
+                            let action: Action = action.into();
+                            let action: Result<TypedAction<EntryCreationData>, WrongActionError> = action.try_into();
+                            let action = action.map_err(|e| wasm_error!(WasmErrorInner::Guest(e.to_string())))?;
+                            match app_entry {}
```
ThetaSinner, line 536: Add a note to this doc comment that this might be simplified once this is released https://github.com/holochain/holochain/pull/5910

src/scaffold/entry_type/integrity.rs
```diff
@@ -547,17 +555,13 @@ fn handle_store_record_arm(
...
-                                    return Ok(ValidateCallbackResult::Invalid(
-                                        "Original action for an update must be a Create or Update action".to_string()
-                                    ));
-                                }
-                            };
+                            let original_action: Result<TypedAction<EntryCreationData>, WrongActionError> = original_action.try_into();
+                            let original_action = original_action.map_err(|e| wasm_error!(WasmErrorInner::Guest(e.to_string())))?;
+                            let creation_action: Action = action.clone().into();
+                            let creation_action: Result<TypedAction<EntryCreationData>, WrongActionError> = creation_action.try_into();
+                            let creation_action = creation_action.map_err(|e| wasm_error!(WasmErrorInner::Guest(e.to_string())))?;
```
ThetaSinner, line 567: Same with this, the conversion would be simplified by https://github.com/holochain/holochain/pull/5910

src/scaffold/entry_type/integrity.rs
```diff
@@ -669,44 +666,34 @@ fn handle_store_entry_arm(
...
                                 validate_create_{snake_entry_def_name}(
-                                    EntryCreationAction::Create(action),
+                                    create_action,
                                     {snake_entry_def_name},
                                 )
                         }},"#
                     ))?;
                     entry_type_match.arms.push(new_arm);
                 }
-            } else if is_entry(&op_entry_arm.pat, "UpdateEntry") {
```
ThetaSinner, line 691: It's not clear to me why this block can be removed?

CLAUDE.md
```diff
@@ -0,0 +1,224 @@
...
+  bumping versions.
+- **Extra `holochain` crate features on a reference app**: this tool only
+  ever enables a small, fixed feature set on the `holochain` dev-dependency
+  (see `initial_cargo_toml()` in `src/scaffold/zome/coordinator.rs` for the
+  current list). If a reference app's `Cargo.toml` has more features
+  enabled at the workspace level than that, it's likely a hand-edit made
+  after scaffolding, not something this tool generated — don't replicate it
+  here just because a reference app has it.
+
+## 6. "It compiled yesterday" — crates.io drift during active RC cycles
```
ThetaSinner: Drop this section, it's not useful to include in general - just deal with it when it happens.

CLAUDE.md
```diff
@@ -0,0 +1,224 @@
...
+Re-run the full §4 verification loop after any such bump — don't assume it's
+safe just because the source diff was empty; confirm `cargo check --workspace
+--all-targets` actually passes.
+
+## 7. If something looks like an upstream regression
+
+If the target release seems to have dropped a helper or introduced a logic
+gap with no replacement (not just a rename), don't paper over it with a
+workaround — flag it and let the user decide. This is expected during RC
+cycles; upstream is often still actively fixing things.
```
ThetaSinner: This bit is good, let's find another home for this in the doc

src/scaffold/entry_type/integrity.rs
ThetaSinner: Added — noted that this two-step conversion may be simplified once holochain/holochain#5910 lands. Applied the same note to all six occurrences of this pattern in the file for consistency (the other four are in this function's UpdateEntry/DeleteEntry arms, `handle_store_entry_arm`'s merged arm, `handle_register_update_arm`, and `handle_register_delete_arm`).

src/scaffold/entry_type/integrity.rs
ThetaSinner: Same fix applied here as the comment above — this note is now on all six occurrences of the Action -> TypedAction conversion pattern in the file.

src/scaffold/entry_type/integrity.rs
ThetaSinner: Both arms had byte-identical generated bodies: under the current hdi API, `OpEntry::CreateEntry` and `OpEntry::UpdateEntry` (both inside `FlatOp::CreateEntry`, formerly `StoreEntry`) narrow the action to the same `TypedAction<EntryCreationData>` and call the same `validate_create_*` function — the entry payload gets identical structural validation whether it's a create or an update-as-create-entry op. This matched what a real hand-upgraded reference app generated, and I confirmed it compiles correctly against the actual hdi crate. So the two branches collapsed into one `is_entry(..., "CreateEntry") || is_entry(..., "UpdateEntry")` condition rather than staying as near-duplicate branches. Nothing was dropped, just deduplicated.

CLAUDE.md
ThetaSinner: Dropped.

CLAUDE.md
ThetaSinner: Moved into the intro, right after the first paragraph — it's a general operating principle for the whole upgrade process rather than a narrow technical detail, so it reads better stated up front.

CLAUDE.md
```diff
@@ -0,0 +1,224 @@
...
+- `HDI_VERSION` (crates.io `hdi`)
+- `HDK_VERSION` (crates.io `hdk`)
+- `HOLOCHAIN_CLIENT_VERSION` (npm `@holochain/client`)
+- `HC_SPIN_VERSION` (npm `@holochain/hc-spin`)
+
+Find the matching set for a target Holochain release from the
+`holochain/holochain` GitHub repo tags (`holochain-X.Y.Z`, `hdi-X.Y.Z`,
+`hdk-X.Y.Z`) or crates.io/npmjs listings — the four crates are released
+together but with independent version numbers, so don't assume they move in
+lockstep.
```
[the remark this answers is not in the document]
ThetaSinner: Valid catch — fixed. Reworded to reflect what's actually there: 3 Rust crates (verified from GitHub tags) + 2 npm packages (verified from npmjs), rather than "four crates."

CLAUDE.md
```diff
@@ -0,0 +1,224 @@
...
+  — feature names do get renamed or removed between releases, so a feature
+  list that compiled against the old target version isn't guaranteed to
+  still be valid.
+
+None of the above four files have unit tests that exercise the generated
+`validate()`/`signal_action` content end-to-end (the existing
+`src/templates/entry_type/tests/*` and `src/templates/link_type/tests/*`
+tests only cover the *coordinator test file* templates, not the integrity
+zome). **Always verify by actually running the CLI**, not by trusting
+`cargo test` — see §4.
```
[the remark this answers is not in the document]
ThetaSinner: Valid catch — fixed, both occurrences now say "five files" (matching the five bullets in section 2, including `zome/coordinator.rs`).

CLAUDE.md
```diff
@@ -0,0 +1,224 @@
...
+# subsequent-arm-push both need coverage.
+"$BIN" -t headless entry-type dino --dna testapp --zome testapp_integrity \
+  --crud crud --reference-entry-hash false \
+  --link-from-original-to-each-update false \
+  --fields "name:String:TextField" --no-ui
+"$BIN" -t headless entry-type nest --dna testapp --zome testapp_integrity \
+  --crud cr --reference-entry-hash false \
+  --fields "dino_hash:ActionHash::Dino" --no-ui
+"$BIN" -t headless link-type nest dino --dna testapp --zome testapp_integrity \
+  --bidirectional false --delete false --no-ui
```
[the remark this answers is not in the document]
ThetaSinner: Valid catch — added a second link-type invocation with `--delete true`, matching the "one deletable, one not" prose. Also swapped one entry-type example to `--reference-entry-hash true`, since that path had an actual bug (see the fixup commit) that a narrower manual check had missed.

src/scaffold/link_type/integrity.rs
```diff
@@ -485,138 +476,138 @@ fn add_link_type_to_validation_arms(
...
-                                            link_type_match.arms.push(new_arm);
-                                        }
-                                    } else if path_segment.ident == "RegisterDeleteLink" {
-                                        // Add new link type to match arm
-                                        if find_ending_match_expr(&mut arm.body).is_none() {
-                                            // Change empty invalid to match on link_type
-                                            *arm.body =
-                                                syn::parse_str::<syn::Expr>("match link_type {}")?;
-                                        }
+                    )?;
```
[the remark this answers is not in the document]
ThetaSinner: Not applying this one. I checked the actual hdi source (`crates/holochain_integrity_types/src/record.rs` at `hdi-0.8.0-rc.3`): `TypedAction<D>` is `pub struct TypedAction<D> { pub header: ActionHeader, pub data: D }` — both fields are public, and there is no `new()` constructor anywhere in the crate. The struct literal is the correct, idiomatic construction here, and I've confirmed this exact code compiles cleanly against the real `hdi 0.8.0-rc.3` crate (not just `cargo check` on this tool, but the generated zome code itself). This looks like a hallucinated API — `TypedAction::new(...)` doesn't exist.

src/versions.rs
```diff
@@ -1,14 +1,14 @@
...
-pub const HDI_VERSION: &str = "0.8.0-dev.11";
+pub const HDI_VERSION: &str = "0.8.0-rc.3";
 
 /// crates.io <https://crates.io/crates/hdk/versions>
-pub const HDK_VERSION: &str = "0.7.0-dev.15";
+pub const HDK_VERSION: &str = "0.7.0-rc.3";
 
 /// crates.io <https://crates.io/crates/holochain/versions>
-pub const HOLOCHAIN_VERSION: &str = "0.7.0-dev.23";
+pub const HOLOCHAIN_VERSION: &str = "0.7.0-rc.4";
```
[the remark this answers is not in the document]
ThetaSinner: Not applying this — it recommends reverting to versions that either don't exist or undo already-verified work. I checked the npm registry directly (`registry.npmjs.org/@holochain/client` and `/@holochain/hc-spin`): `0.21.0-rc.1` and `0.700.0-rc.1` are both published, contradicting the claim they aren't. Likewise `hdi=0.8.0-rc.3`/`hdk=0.7.0-rc.3`/`holochain=0.7.0-rc.4` are real and published — this is the exact combination currently pinned, verified by `cargo generate-lockfile` resolving cleanly and the full workspace (including the sweettest dev-dependency, which needed exactly this bump to resolve a crates.io drift issue) compiling against them. The review's own linked tool output shows failed/empty script runs before it falls back to web-query results that don't match the real registries — this suggestion is based on hallucinated data.

ThetaSinner:
Addressing CodeRabbit's three "outside diff range" findings (not inline-replyable, so posting here):

**`src/scaffold/entry_type/integrity.rs` (601-616, 762-779) — "use `original_action.data.entry_type()` instead of `original_action.entry_type()`"**: Not applying. `TypedAction<EntryCreationData>` has an inherent `entry_type()` method (`crates/holochain_integrity_types/src/typed_action.rs` at `hdi-0.8.0-rc.3`: `impl TypedAction<EntryCreationData> { pub fn entry_type(&self) -> &EntryType { self.data.entry_type() } }`), so `original_action.entry_type()` is correct as written. Confirmed by actually compiling this exact code against the real `hdi 0.8.0-rc.3` crate — not just checked, compiled clean.

**`src/scaffold/entry_type/integrity.rs` (221-256) — "hdi 0.8.0-rc.3 doesn't expose `TypedAction`/`EntryCreationData`, use legacy `EntryCreationAction`"**: Not applying — this is backwards. `EntryCreationAction` is the *old* API (from `0.7.0-dev.23`) that this PR specifically migrates away from. `TypedAction<EntryCreationData>` is the real, current type, confirmed by direct inspection of the hdi source and by compiling the generated zome code against the actual published crate.

**`src/scaffold/zome/integrity.rs` (62-100) — "add a fallback arm for `StoreRecord`/`StoreEntry`/`RegisterUpdate`/etc."**: Not applying. Those are the *old* `FlatOp` variant names this PR renames away from. The current `hdi::flat_op::FlatOp` enum has exactly six variants — `CreateRecord`, `CreateEntry`, `AgentActivity`, `Link`, `Update`, `Delete` — all of which are matched, confirmed by the fact this compiles without a non-exhaustive-match error against the real crate (Rust would reject an incomplete match otherwise).

All three read as hallucinated from a stale/partial model of the hdi API (the CodeRabbit review's own linked analysis shows failed sandbox script runs before falling back to web-query results that turned out to be wrong elsewhere in this review too — see the inline reply on the `versions.rs` and `TypedAction::new()` comments). Happy to be shown otherwise with a link to real hdi source at the target tag, but I've verified end-to-end (including a full `run_test.sh`-equivalent scaffold + `cargo check --workspace --all-targets` + `cargo clippy`) and everything compiles clean.

